PR #3, opened 2019-02-11
Switch to pytest for running unittests.  Clean up some warnings.

iloveagent57:
## Description
* Removes nose in favor of pytest.
* Cleans up some Django deprecations warnings.
* https://openedx.atlassian.net/browse/EDUCATOR-4041

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ create_superuser:
 	docker exec -t registrar-app bash -c 'echo "from django.contrib.auth import get_user_model; User = get_user_model(); User.objects.create_superuser(\"edx\", \"edx@example.com\",\"edx\")" | python manage.py shell'
 
 
-# The followeing targets must be build from within the Docker container,
+# The followeing targets must be built from within the Docker container,
 # which can be accessed using `make shell` after running `make up`.
 
 upgrade: piptools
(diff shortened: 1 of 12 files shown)


schenedx (approved):
👍

Makefile
```diff
@@ -97,9 +97,12 @@ requirements:
 prod-requirements:
 	pip-sync -q requirements.txt
 
+coverage: clean
```
kdmccormick, line 100: This will make it so `make coverage` re-runs tests. Maybe have the following setup instead?
```Makefile
coverage: htmlcov/index.html
    $(BROWSER) htmlcov/index.html

htmlcov/index.html: test

test:
    pytest --cov-report html
```

Makefile
```diff
@@ -109,7 +112,7 @@ pii_check:
 	DJANGO_SETTINGS_MODULE=registrar.settings.test \
 	code_annotations django_find_annotations --config_file .pii_annotations.yml --lint --report --coverage
 
-validate: test quality pii_check
+validate: coverage quality pii_check
```
kdmccormick, line 115: If you choose to take my previous comment, then this would become `test` again.

kdmccormick (approved):
Left a comment. Whether or not you choose to take it, :+1:

Makefile
iloveagent57: The `coverage` target is really meant to be run from a CI context, where you might want to stash some HTML coverage reports somewhere to access via a UI.  And we want to make sure that the directive actually runs test before generating coverage.

For local dev, you'll run `make test`, which will also spit out a coverage report due to the options included in `pytest.ini`.

kdmccormick: Gotcha. Everything LGTM then.

PR #4, opened 2019-02-11
Remove a couple deprecation warnings

kdmccormick:
https://openedx.atlassian.net/browse/EDUCATOR-4043

### DIFF
--- a/registrar/apps/api/urls.py
+++ b/registrar/apps/api/urls.py
@@ -6,8 +6,9 @@ contain namespaces for the active versions of the API.
 """
 from django.conf.urls import url, include
 
+from registrar.apps.api.v1 import urls as v1_urls
 
 app_name = 'api'
 urlpatterns = [
-    url(r'^v1/', include('registrar.apps.api.v1.urls')),
+    url(r'^v1/', include(v1_urls)),
 ]
--- a/registrar/settings/base.py
+++ b/registrar/settings/base.py
@@ -158,7 +158,7 @@ LOGOUT_URL = '/logout/'
 AUTH_USER_MODEL = 'core.User'
 
 AUTHENTICATION_BACKENDS = (
-    'auth_backends.backends.EdXOpenIdConnect',
+    'auth_backends.backends.EdXOAuth2',
     'django.contrib.auth.backends.ModelBackend',
 )
 
--- a/registrar/urls.py
+++ b/registrar/urls.py
@@ -21,16 +21,19 @@ from django.conf.urls import include, url
 from django.contrib import admin
 from rest_framework_swagger.views import get_swagger_view
 
+from registrar.apps.api import urls as api_urls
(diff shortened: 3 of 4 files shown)


registrar/apps/api/urls.py
```diff
@@ -6,8 +6,9 @@
 """
 from django.conf.urls import url, include
 
+from registrar.apps.api.v1 import urls as v1_urls
 
 app_name = 'api'
 urlpatterns = [
-    url(r'^v1/', include('registrar.apps.api.v1.urls')),
+    url(r'^v1/', include(v1_urls)),
```
kdmccormick, line 13: This is a personal style preference -- I like using actual modules instead of module names, because if improperly configured they'll break in tests and not just when you try to hit the URL. If you disagree  @iloveagent57 I'll change them back.

registrar/settings/base.py
```diff
@@ -158,7 +158,7 @@
 AUTH_USER_MODEL = 'core.User'
 
 AUTHENTICATION_BACKENDS = (
-    'auth_backends.backends.EdXOpenIdConnect',
+    'auth_backends.backends.EdXOAuth2',
```
kdmccormick, line 161: I have no idea if there's more to be done here. I think we'll have to actually try authenticating against an LMS to be sure.

registrar/urls.py
```diff
@@ -21,16 +21,19 @@
...
+from registrar.apps.api import urls as api_urls
 from registrar.apps.core import views as core_views
 
 admin.autodiscover()
 
+app_name = 'registrar'
+
 urlpatterns = auth_urlpatterns + [
     url(r'^admin/', admin.site.urls),
-    url(r'^api/', include('registrar.apps.api.urls', namespace='api')),
```
kdmccormick, line 30: I don't think these `namespace` parameters are necessary.

registrar/urls.py
```diff
@@ -21,16 +21,19 @@
 from django.contrib import admin
 from rest_framework_swagger.views import get_swagger_view
 
+from registrar.apps.api import urls as api_urls
 from registrar.apps.core import views as core_views
 
 admin.autodiscover()
 
+app_name = 'registrar'
```
kdmccormick, line 29: This is probably the sensible app name?

registrar/settings/base.py
iloveagent57: Haha, good call, I read that warning as something in auth_backends that needed updating.

registrar/apps/api/urls.py
iloveagent57: That's fine with me 👍

registrar/urls.py
iloveagent57: Yes, good catch.

iloveagent57 (approved):

PR #8, opened 2019-02-12
ADR for creation of a registrar service.

iloveagent57:
## Description

We need an ADR that captures the motivating factors and consequences of creating a registrar service.

### DIFF
--- a/docs/decisions/0001-why-we-need-registrar.rst
+++ b/docs/decisions/0001-why-we-need-registrar.rst
@@ -1,0 +1,85 @@
+Why we need a Registrar service
+===============================
+
+Context
+=======
+
+The viability of Master’s programs on edX will be partly determined by the extent to which schools
+can scale admissions beyond what they can currently handle on campus.  This needs to be supported by edX
+reducing operational friction to a minimum, both at the schools and at edX.
+In order to affect this we need to automate as many operational procedures as possible,
+and eliminate wherever possible the need for any manual procedures by either school or edX staff.
+Different schools may manage their program enrollees differently,
+but we want to provide a common interface for the management of program enrollees on edX.
+This will ensure that we can provide a consistent experience for all Master's students
+and provide a common set of data and integration points to all of our partners.
+The following points of integration have been identified:
+
+#. Program admission
+#. Course enrollment
+#. Publishing of final course grades
+#. Automated creation of course shells on edX
+#. Retrieve course attendance
+
+Decision
+========
+
+We will introduce a new service called "registrar".  The primary purpose of this service is
+the management of data related to a learner's enrollment in Master's (and eventually other types)
+programs and courses therein.
+
+- This service will explicitly persist student-program admission records (i.e. it
+  records the fact that a student is enrolled in a program with some status).
+- It will identify the courses in which the admitted student is enrolled in the
+  pursuit of satisfying a program's requirements.
+- It will serve as the primary integration point between partner Student Information Systems ("SIS")
+  and edX systems.
+- It will NOT explicitly capture the ways in which a program's requirements are satisfied
+  (this is left to the partner's SIS).
+- This service will communicate and synchronize data with other edX services
+  via REST APIs.
+
+  - Our initial synchronization strategy will likely by an ad-hoc strategy - we'll refresh
+    data from other systems as needed, if it has not recently been refreshed (or if
+    it is not currently persisted in registrar).
+
+Status
+======
+
+Accepted (circa February 2019)
+
+Consequences
+============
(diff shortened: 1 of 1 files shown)


jansenk (approved):

docs/decisions/0001-why-we-need-registrar.rst
```diff
@@ -0,0 +1,81 @@
...
+===============================
+
+Context
+=======
+
+The viability of Master’s programs on edX will be partly determined by the extent to which schools
+can scale admissions beyond what they can currently handle on campus.  This needs to be supported by edX
+reducing operational friction to a minimum, both at the schools and at edX.
+In order to affect this we need to automate as many operational procedures as possible,
+and eliminate wherever possible the need for any manual procedures by either school or edX staff.
```
schenedx, line 11: To me, another driving factor for this is to establish a single set of rules of engagement between the schools and edX to collaboratively administer Master programs. Different school may manage their Program students differently, but we need to make sure they have a common set of interface when they manage students on edX platform.

schenedx (approved):
Just a minor context addition. Feel free to merge as it is, or take in my suggestion

docs/decisions/0001-why-we-need-registrar.rst
iloveagent57: I've updated this section to include this factor.

docs/decisions/0001-why-we-need-registrar.rst
```diff
@@ -0,0 +1,85 @@
...
+Different school may manage their program enrollees differently,
```
nasthagiri: nit: school -> schools

docs/decisions/0001-why-we-need-registrar.rst
```diff
@@ -0,0 +1,85 @@
...
+  with the edX system.
+
+  - Rather than having integrate with the course-discovery service for some data needs
+    and the LMS for other data needs, all program data integration can be done through one
+    service, and the API specification to support third party needs can be captured
+    within a single service and schema.
+
+- We'll have explicit records representing the relationship between learners and (Master's) programs.
+
+  - This will greatly simplify our the ease with which we can reason about a learner's relationship
```
nasthagiri: Nit:  our the -> the

docs/decisions/0001-why-we-need-registrar.rst
```diff
@@ -0,0 +1,85 @@
+Why we need a Registrar service
```
nasthagiri, line 1: Thanks for writing up this ADR and description of the service.  I hope it will help clarify to the team and to future developers where this new service fits in, per its "Single Responsibility".

docs/decisions/0001-why-we-need-registrar.rst
```diff
@@ -0,0 +1,85 @@
...
+#. Program admission
+#. Course enrollment
+#. Publishing of final course grade
+#. Automated creation of course shells on edX
+#. Retrieve course attendance
+
+Decision
+========
+
+We will introduce a new service called "registrar".  The primary purpose of this service is
```
nasthagiri, line 27: According to the [dictionary](https://www.dictionary.com/browse/registrar), a `registrar` is:

- `an official in a college or university who is responsible for keeping student records`
- `an official at a school or college who maintains students' personal and academic records, issues reports of grades, mails out official publications, etc.`

From that perspective, I'd like to know more about the naming choice - since "registrar" implied an entity that maintains student records - rather than student "enrollments".   Have you gotten input from Nina on this naming choice?

iloveagent57: Good question, the Neem team had a discussion about this today.  There's a strong desire (from both the engineering team and product team) to not use the word "enrollments" here, since it's usage would be overloaded w.r.t. other usages in edX systems.  We discussed "admissions", but that word feels closer to something that determines whether a student is admitted, which is not really the purpose here.

Also, while this system (edx/registrar) isn't tasked with persisting student grade records, it is responsible for reporting those grades back to partner institutions.  There's also an argument to be made that the recording of a student's status w.r.t. a given degree (i.e. "pending", "active", "completed") is an academic record, and analogous to a physical registrar's office record of the same type.

I'm happy to double check with Nina or Melanie, too.

nasthagiri: Not intending to add more work here.  This ADR in itself is great since it provides the purpose of this service and repo.  If others are misled by the name or want to know deeper about its intent, they can refer to this write-up.  Thanks again!

nasthagiri (approved):

PR #11, opened 2019-02-25
Various changes to Organization and Program models

kdmccormick:
**Team:** @edx/educator-neem

Because we will be using organization-assigned student IDs for enrolling learners in a program, every program must have a single managing organization. So, we delete the many-to-many field `Program.organizations` and change it to the one-to-many field `Program.manging_organization`. If it becomes the case in the future that program can have multiple authoring organizations, we can add a separate many-to-many field to `Program`.

Also, we add a `key` field to `Program` that will be equivalent to the `marketing_slug` from discovery, and add a uniqueness constraint to `Organization.key` so that it's guaranteed that every organization in the database can be identified.

Finally, we create the `Program.url` field.


iloveagent57 (approved):

schenedx (approved):
👍

PR #15, opened 2019-02-27
Add ADR 0002 Keeping the Program Enrollment Source of Truth within Registrar

schenedx:
## Description

This is the ADR on keeping the Program Enrollment Source of Truth within Registrar

@edx/educator-neem Please review and comment

### DIFF
--- a/docs/decisions/0002-keeping-program-enrollment.rst
+++ b/docs/decisions/0002-keeping-program-enrollment.rst
@@ -1,0 +1,47 @@
+Keeping the Program Enrollment Source of Truth within Registrar
+===============================================================
+
+Context
+=======
+
+Question have been raised by the edX architecture team about the fact that we have Program Enrollment
+explicitly stored in Registrar service. They have the following concerns:
+
+*. Registrar service seems to be servicing the function of being an integration service between
+   school's system with edX about program administration
+*. Program Enrollment is an edX domain concept that would be used by edX specific components to 
+   realize functionality such as learner program progress display.
+*. If Program enrollment data is needed outside of Registrar from services like LMS or Credentials,
+   the synchronization of the critical data might become a production issue difficult to resolve
+   and maintain
+
+
+Decision
+========
+
+After careful consideration, we made the decision to keep Program Enrollment table within 
+Registrar service. The decision is made based on the following:
+
+- We believe this is a reversible decision within the second half of FY19.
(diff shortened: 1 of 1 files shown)


kdmccormick (changes requested):
Looks generally good. I'd prefer, though, if there were a bullet point clarifying that we do **not** want Registrar to serve as a source of truth for non-Master's program enrollments at this time, and that if that there becomes a need to store general program enrollment data, creating a non-Registrar solution should be considered.

schenedx:
@kdmccormick Took in your feedback. Please review again

kdmccormick (approved):

kdmccormick:
Thanks! LGTM :+1:

docs/decisions/0002-keeping-program-enrollment.rst
```diff
@@ -0,0 +1,47 @@
...
+*. Registrar service seems to be servicing the function of being an integration service between
+   school's system with edX about program administration
+*. Program Enrollment is an edX domain concept that would be used by edX specific components to 
+   realize functionality such as learner program progress display.
+*. If Program enrollment data is needed outside of Registrar from services like LMS or Credentials,
+   the synchronization of the critical data might become a production issue difficult to resolve
+   and maintain
+
+
+Decision
```
nasthagiri, line 19: Here are my thoughts on this ADR:
1. Reversibility of a decision is not binary.  It’s **qualified** by the amount of effort to _actually_ reverse and clean up afterward.
2. The primary arch input is on the conceptual models/naming/boundaries and system-wide data flow:
   2a. Mutually understanding that the Registrar is an intermediary integration service between external schools and the rest of the platform.
   2b. The data flow is unidirectional from the Registrar to the platform.  We will not accept any code that calls from the platform to this new service.  If the team expects there to be a need for bi-directional data flow, then we need to reverse this decision right now.
   2c. Since Registrar is its own bounded context, it can choose to store any context-specific data within it own database.  In order to avoid confusion, we recommend **not** naming data within the Registar as “enrollments” since that is a platform concept.
3. From the description of the rationale given in this section, it seems we are moving forward with a decision that we agree is _not_ the correct direction. In this situtation, we continue to _strongly_ recommend doing what we believe is the right path today, rather than waiting and revisting after several months.  Regardless, we would like to be invited to a calendar invite for a future date to revisit this, so it doesn't get forgotten forever.

PR #17, opened 2019-03-06
add guardian

jansenk:
Add guardian to registrar requirements and settings

make Org admin inherit from GuardedModelAdmin

### DIFF
--- a/registrar/settings/base.py
+++ b/registrar/settings/base.py
@@ -27,6 +27,7 @@ INSTALLED_APPS = (
 )
 
 THIRD_PARTY_APPS = (
+    'guardian',
     'rest_framework',
     'rest_framework_swagger',
     'simple_history',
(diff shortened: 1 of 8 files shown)


registrar/settings/base.py
```diff
@@ -24,6 +24,7 @@
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
+    'guardian',
```
iloveagent57: I think this really belongs in `THIRD_PARTY_APPS` below.

iloveagent57 (commented):
Just a one line change, everything else looks good

iloveagent57:
Ah, you'll also need to add some pii_checks for the guardian models.

iloveagent57 (approved):

PR #21, opened 2019-03-08
Add OrgGroups and roles

jansenk:
https://openedx.atlassian.net/browse/EDUCATOR-4151

Add OrgGroups
Add Organization permissions and roles
Tests


zacharis278 (approved):
lgtm

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -11,7 +11,7 @@
     StaffUserFactory,
 )
 from registrar.apps.enrollments.models import Organization
-from registrar.apps.enrollments.tests.factories import (
+from registrar.apps.enrollments.tests.factories import (  # pylint: disable=no-name-in-module
```
kdmccormick: pylint passes for me without these `no-name-in-module` suppressions. Was pylint failing locally for you, or on jenkins, or both?

jansenk: It was failing for me locally

registrar/apps/enrollments/models.py
```diff
@@ -45,6 +55,41 @@ def __str__(self):
...
+        app_label = 'enrollments'
+        verbose_name = 'Organization Group'
+
+    ROLE_CHOICES = (
+        (perms.OrganizationReadMetadataRole.name, 'Read an Organization Metadata'),
+        (perms.OrganizationReadEnrollmentsRole.name, 'Read Enrollments Data'),
+        (perms.OrganizationReadWriteEnrollmentsRole.name, 'Read and Write Enrollments Data'),
+    )
+
+    organization = models.ForeignKey(Organization, on_delete=models.CASCADE)
```
kdmccormick, line 74: If we're doing `organization = None` for global permissions, we need `null=True` here, along with a comment saying so.

Additionally, `OrgGroup.save` will need to be tweaked, and we should have a test verifying that `organization = None` implies global permissions.

registrar/apps/api/v1/tests/test_views.py
kdmccormick, line 14: pylint passes for me without these `no-name-in-module` suppressions. Was pylint failing locally for you, or on jenkins, or both?

kdmccormick (changes requested):
Looks good other than the comments I left.

registrar/apps/enrollments/models.py
```diff
@@ -45,6 +55,41 @@ def __str__(self):
         return self.name
 
 
+class OrgGroup(Group):
```
kdmccormick: If there's a way to remove the `Permissions` dialog from the admin panel, that'd be a good thing to do.

registrar/apps/enrollments/models.py
```diff
@@ -45,6 +55,41 @@ def __str__(self):
...
+    Group subclass to grant select guardian permissions to a group on an organization level
+
+    .. no_pii::
+    """
+    class Meta(object):
+        app_label = 'enrollments'
+        verbose_name = 'Organization Group'
+
+    ROLE_CHOICES = (
+        (perms.OrganizationReadMetadataRole.name, 'Read an Organization Metadata'),
```
kdmccormick: I recommend changing this to `'Read Metadata Only'`, because:
* It's program metadata they're reading, not organization metadata
* Having `'Only'` in the name hints that the other two roles allow metadata reading too

registrar/apps/enrollments/models.py
schenedx: For the statement:
`If we're doing organization = None for global permissions, we need null=True here, along with a comment saying so.`, I thought we just won't have a data row into this table if the group created is not limited to any organization, right?

iloveagent57: That's right, Simon - we'll use vanilla `auth_group` records for global perms - every `OrgGroup` should have a non-null organization.

registrar/apps/enrollments/models.py
iloveagent57: Let the great debate of our time begin: `OrgGroup`, `OrganizationGroup`, or something else?  `AccessControlGroup`?  Also, does this belong in the `enrollments` app or the `core` app?  I think I can make a pretty good case that it belongs in `core`.

registrar/apps/enrollments/tests/test_models.py
```diff
@@ -0,0 +1,52 @@
...
+        super(OrgGroupTests, self).setUp()
+        self.organization = OrganizationFactory()
+        self.user = UserFactory()
+
+    @ddt.data(
+        perm.OrganizationReadMetadataRole,
+        perm.OrganizationReadEnrollmentsRole,
+        perm.OrganizationReadWriteEnrollmentsRole,
+    )
+    def test_roles(self, role):
```
schenedx, line 28: Can we have another test to specifically make sure, if a user is created and saved into org 1, it do not have permissions for any permissions on org 2?

schenedx (commented):
suggested one more test case.

registrar/apps/enrollments/models.py
schenedx: I suggest `core` as well. I can see we have different django app for different data set to proxy with

registrar/apps/api/v1/tests/test_views.py
jansenk: Yeah, this was failing for me locally

registrar/apps/enrollments/models.py
kdmccormick: My vote is `OrganizationGroup`. Unnecessary abbreviation is confusing.

I think it belongs in `core`, but then `Organization` should also be moved to `core`. `core` shouldn't be importing from `enrollments`.

registrar/apps/enrollments/models.py
kdmccormick: Sounds good; my misunderstanding.

registrar/apps/enrollments/models.py
schenedx: @kdmccormick I agree we should move Organization out of enrollment app. 
Can we just create a new ticket to do the move instead of adding scope to this PR? @jansenk do you mind creating the ticket to capture the conversation above?

schenedx: @jansenk I already created the move ticket at https://openedx.atlassian.net/browse/EDUCATOR-4184

schenedx (approved):
👍

PR #23, opened 2019-03-11
Add ADR for AuthN and AuthZ schemes.

iloveagent57:
## Description

* Adds an ADR to document our recent decisions around AuthN/Z.
* I've tried to document the precise decisions we've made without going too far into implementation details - the referenced resources contain more details if the reader is interested.
* Please point out any other consequences you think are worth mentioning, or other decision points that should be documented.

### DIFF
--- a/docs/decisions/0003-authnz.rst
+++ b/docs/decisions/0003-authnz.rst
@@ -1,0 +1,68 @@
+Authentication and Authorization
+================================
+
+Context
+-------
+
+External API clients of the Registrar service will need to be authenticated (AuthN) and authorized (AuthZ)
+to access the resources they request.  An API client will usually correspond to a human associated
+with a Master's partner institution (school) who will make HTTP requests of registrar.  The purpose
+of these requests is to either read data about resources, create resources, or update resources.
+
+Decision
+--------
+
+Authentication
+~~~~~~~~~~~~~~
+
+1. We'll use the LMS as the identity provider for registrar.  It will provide a JSON Web Token (JWT)
+   to authenticated users.
+2. The ``edx-drf-extensions`` library provides an authentication class ``JwtAuthentication``
+   which will facilitate the creation of a ``core_user`` account in registrar based on the user profile
+   information provided in the JWT.
+3. The LMS endpoint ``/api-admin/`` will allow API clients of registrar to manage their own OAuth2 client credentials.
+   See https://github.com/edx/edx-platform/tree/master/openedx/core/djangoapps/api_admin
+4. We'll also allow authenticated clients to access the DRF browsable API.
+
+Authorization
+~~~~~~~~~~~~~
+
+1. We'll implement a lightweight version of role-based access control (RBAC) of registrar API resources.
+2. Every API client associated with a school will be a member of one or more "Access Control" groups.  These
+   groups are extended Django groups that we can associate an ``Organization`` with.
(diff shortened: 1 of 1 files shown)


jansenk (approved):

docs/decisions/0003-authnz.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Authentication
+--------------
+
+1. We'll use the LMS as the identity provider for registrar.  It will provide a JSON Web Token (JWT)
+   to authenticated users.
+2. The ``edx-drf-extensions`` library provides an authentication class ``JwtAuthentication``
+   which will facilitate the creation of a ``core_user`` account in registrar based on the user profile
+   information provided in the JWT.
+3. The LMS endpoint ``/api-admin/`` will allow API clients of registrar to manage their own OAuth2 client credentials.
```
schenedx, line 23: Is the api-admin flow described somewhere? Can you link that to here?

docs/decisions/0003-authnz.rst
```diff
@@ -0,0 +1,67 @@
...
+1. In the context of global permission-checking and Guardian's ``PermissionRequiredMixin`` for view-set classes,
+   it's going to be easier if we break out retrieval actions and list actions
+   (i.e. get one program vs. get many programs) into separate view sets - the logic for fetching an
+   ``Organization`` object in the context of a retrieval action (where we're passed a program key) vs.
+   fetching an ``Organization`` object during a list action (where we're passed either an org key or nothing at all)
+   is sufficiently different to warrant this.
+2. We know of at least one school with two programs - call them "A" and "B" - where administrators of program
+   A should not be able to access resources from program B, and vica versa.  We have two options here:
+
+   a. Create distinct ``Organizations`` for "A" and "B".  The downside here is that we lose the ability to
```
schenedx, line 59: Should we made a decision here? Or that will come in a later ADR?

schenedx (commented):
I have a question and a suggestion.

docs/decisions/0003-authnz.rst
```diff
@@ -0,0 +1,67 @@
+Authentication and Authorization
+================================
```
kdmccormick, line 2: Nit: Move all headings down one "level" so that this is the title, and the rest are sections. So `=====` becomes `-----` and `-----` becomes `~~~~~`.

kdmccormick (approved):
One minor nit. LGTM.

Question relevant to this: If we are checking whether a user should be able to do "elevated" actions (such as listing programs across all orgs), should we check `user.is_staff` or `user.is_superuser`?

docs/decisions/0003-authnz.rst
iloveagent57: It's described in the confluence doc.  I can add a link to the code, at least.

docs/decisions/0003-authnz.rst
iloveagent57: Good point.

On your other question: `is_staff` indicates that the user can access the admin site, and `is_superuser` indicates that the user has every possible permission.  I think it's valid to have internal edX users who can't access the admin site but should be able to access programs across all orgs, so we can just do the global `user.has_perm(permission_name)` check without worry about their admin site access.

docs/decisions/0003-authnz.rst
iloveagent57: I think that's more of an implementation-level decision that we can capture elsewhere.

PR #24, opened 2019-03-11
Add ADR #2: Program Enrollment Storage

kdmccormick:


docs/decisions/0002-program-enrollment-storage.rst
```diff
@@ -0,0 +1,73 @@
...
+-------
+
+Currently, the idea of "program enrollments" does not exist explictly anywhere
+in the edX ecosystem. When students purchase enrollment in a program today, they
+are implicitly enrolled by being given an Entitlement
+(i.e., purchased seat) in every course within the program.
+However, there is no table anywhere that links together students and programs,
+like there is for students and courses. For Master's degree programs, this implicit
+enrollment system will not suffice.
+While is clear that a ProgramEnrollment table must created, there has been much
```
reillz10: "While is clear" -> "While it is clear"

reillz10: also, NIT the word "much" doesn't really do much to enhance understanding here.

docs/decisions/0002-program-enrollment-storage.rst
```diff
@@ -0,0 +1,73 @@
...
+account is created, so that any corresponding preemptive enrollments can be activated.
+It would be simpler to implement program enrollments, both preemptive and actual,
+within the LMS.
+
+Finally, an alternative viewpoint was for creating an "Enrollments Service" to store
+both preemptive and actual program enrollments (and, in the future, course enrollments),
+while having the Registar remain solely an integration point.
+This would both avoid adding features to the LMS as well as avoid adding authoritative
+core data to Registrar.
+There was worry, however, that migrating course enrollments over to such an Enrollments
```
reillz10: NIT: "There was worry, however, that" -> "However, there was worry"

The however is distracting me from the worry and the thing we are worrying about.

docs/decisions/0002-program-enrollment-storage.rst
```diff
@@ -0,0 +1,73 @@
+Storing Program Enrollment within edX Platform
+==============================================
+
+Context
+-------
+
+Currently, the idea of "program enrollments" does not exist explictly anywhere
+in the edX ecosystem. When students purchase enrollment in a program today, they
+are implicitly enrolled by being given an Entitlement
```
schenedx: This is not the implicit mechanism. The entitlement has nothing to do with learner being part of a program. The implicit logic is based on the following conditions:
1. Student got course_run verified cert
2. The course_run is part of the course within a program

And we will award Program cert if the condition below becomes true:
* student got ALL the course cert of a included course_run of a course in a program.

docs/decisions/0002-program-enrollment-storage.rst
```diff
@@ -0,0 +1,73 @@
...
+and are likely to be needed by other edX services and frontends
+(e.g., the Master's Student Dashboard).
+It would be undesirable for other parts of the edX ecosystem to depend on a Registrar,
+an integration point, for authoritative data.
+Furthermore, if program enrollments were stored in Registrar,
+then "preemptive program enrollments"
+(program enrollments made for a students who do not yet have edX accounts)
+would also need to be stored in Registrar.
+This would require the LMS to somehow communicate to Registrar whenever an
+account is created, so that any corresponding preemptive enrollments can be activated.
```
schenedx, line 36: `whenever a learner account is created`

schenedx (commented):
Just a few minor suggestions

docs/decisions/0002-program-enrollment-storage.rst
```diff
@@ -0,0 +1,73 @@
...
+Status
+------
+
+Accepted (circa March 2019)
+
+Consequences
+------------
+
+* The ``ProgramEnrollment`` and ``Learner`` models will be removed from Registrar.
+* The ``ProgramEnrollment``, ``PreemptiveProgramEnrollment``, and ``PreemptiveCourseEnrollment`` models will be created within LMS.
```
iloveagent57: Are we sure we don't want to keep a "cached" version of these models in registrar?  I'd been assuming that we would.
Also, there's already a thing to do future course enrollments: https://github.com/edx/edx-platform/blob/e8555de4b42a8f9b48c174cdac9c992ce3910b30/common/djangoapps/student/models.py#L1994

docs/decisions/0002-program-enrollment-storage.rst
```diff
@@ -0,0 +1,73 @@
...
+it would remove the need for us to tailor the table to Master's business needs,
+instead of taking into consideration all edX Programs.
+
+The opposing view is that the table belongs within the LMS application of the edX Platform.
+Program enrollments should be considered a core edX concept,
+and are likely to be needed by other edX services and frontends
+(e.g., the Master's Student Dashboard).
+It would be undesirable for other parts of the edX ecosystem to depend on a Registrar,
+an integration point, for authoritative data.
+Furthermore, if program enrollments were stored in Registrar,
```
iloveagent57, line 31: To me, this is the best argument for storing enrollment data in the LMS.

docs/decisions/0002-program-enrollment-storage.rst
```diff
@@ -0,0 +1,73 @@
...
+like there is for students and courses. For Master's degree programs, this implicit
+enrollment system will not suffice.
+While is clear that a ProgramEnrollment table must created, there has been much
+debate over exactly where that table belongs.
+
+One view is that the table belongs in the Registrar service.
+This prevents us from introducing additional features into edX Platform,
+which is generally considered to be overly large and complex.
+Additionally, as Registrar is currently planned to only service Master's programs,
+it would remove the need for us to tailor the table to Master's business needs,
```
iloveagent57: s/remove the need/allow/

iloveagent57 (commented):
Couple of minor points.  My main desires for maintaining a "cached" version of data are as follows:
1) If we're confident that enrollment data in registrar is up-to-date, it allows for improved read performance.
2) It allows us to do some bookkeeping - we can easily tell if there's a communication problem b/w registrar and LMS if we have a registrar record without a corresponding LMS record.  It lets us keep track of registrar API client behavior without necessarily being a source of truth.
3) Allows for flexibility around storing data that's pertinent to an external API client but isn't needed in the table of record - e.g. we can stash a school ID on the registrar Learner table, but we can't do this with LMS' `auth_user` table (I know you're planning on storing that school ID on the enrollment record anyway).

docs/decisions/0002-program-enrollment-storage.rst
kdmccormick: are you referring to the "Registrar isn't core" or the "preemptive enrollments would be harder" argument?

docs/decisions/0002-program-enrollment-storage.rst
kdmccormick: I didn't think we'd be caching enrollments. But let's talk about this; you have good points.

That's not my reading of `CourseEnrollmentAllowed`. It seems to be a model for allowing existing users to enroll in a course past the enrollment deadline.

iloveagent57: I'm pretty sure it exists to allow instructors to manually enroll students with some email address who may not have created an LMS account yet, like you can do in the instructor dashboard: https://github.com/edx/edx-platform/blob/afd262e3e1239f3ce30dc313afd39e85e6ce5f81/lms/djangoapps/instructor/enrollment.py

docs/decisions/0002-program-enrollment-storage.rst
iloveagent57: The "preemptive enrollments would be harder" argument.

docs/decisions/0002-program-enrollment-storage.rst
kdmccormick: Per our discussion:
* I changed it to say we may or may not store enrollments in Registrar, but if so, they will not be authoritative.
* I changed it to say we will store preemptive enrollments in the LMS, whether that be through existing infrastructure or new models.

Both these points are things we should circle back to, but don't need to be decided in this ADR.

kdmccormick:
@reillz10 @schenedx @iloveagent57 Thanks all for the comments; ready for re-review.

iloveagent57 (approved):
Looks good!

schenedx (approved):
👍

reillz10 (approved):
👍

PR #26, opened 2019-03-12
Program course listing

zacharis278:
### [EDUCATOR-4092](https://openedx.atlassian.net/browse/EDUCATOR-4092)

### DIFF
--- a/registrar/apps/api/serializers.py
+++ b/registrar/apps/api/serializers.py
@@ -41,3 +41,13 @@ class LearnerProgramEnrollmentSerializer(serializers.Serializer):
     learner = LearnerSerializer()
     program = ProgramSerializer()
     status = serializers.CharField()
+
+
+class CourseRunSerializer(serializers.Serializer):
+    """
+    Serializer for a course run returned from the
+    Course Discovery Service
+    """
+    course_id = serializers.CharField(source='key')
+    course_title = serializers.CharField(source='title')
+    course_url = serializers.CharField(source='marketing_url')
--- a/registrar/apps/api/v1/views.py
+++ b/registrar/apps/api/v1/views.py
@@ -1,15 +1,22 @@
 """
 The public-facing REST API.
 """
-from django.http import HttpResponseForbidden
+import logging
+
+from django.http import HttpResponseForbidden, HttpResponseServerError
 from django.shortcuts import get_object_or_404
+from requests.exceptions import HTTPError
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
 from rest_framework import viewsets
+from rest_framework.decorators import action
 from rest_framework.response import Response
 
 from registrar.apps.api.permissions import program_access_required
-from registrar.apps.api.serializers import ProgramSerializer
+from registrar.apps.api.serializers import ProgramSerializer, CourseRunSerializer
 from registrar.apps.enrollments.models import ACCESS_READ, Program, Organization
+from registrar.apps.enrollments.data import get_discovery_program
+
+logger = logging.getLogger(__name__)
 
 
 class ProgramReadOnlyViewSet(viewsets.ReadOnlyModelViewSet):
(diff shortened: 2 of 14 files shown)


registrar/apps/api/serializers.py
```diff
@@ -41,3 +41,14 @@ class LearnerProgramEnrollmentSerializer(serializers.Serializer):
...
     program = ProgramSerializer()
     status = serializers.CharField()
+
+
+class CourseRunSerializer(serializers.Serializer):
+    """
+    Serializer for a course run returned from the
+    Course Discovery Service
+    """
+    key = serializers.CharField()
```
kdmccormick: Please base these fields names off of [the API spec](https://docs.google.com/document/d/1x3nIpJ16KJelS-OoTxZZgyRAyoBXH6EKVxhH-C6SEgw/edit?ts=5c892a27#heading=h.t83iv1vesc0v). The field names on the underlying object can be specified with `source`

registrar/apps/api/v1/views.py
```diff
@@ -57,3 +64,27 @@ def list(self, request):
...
+        service.
+        """
+        program = self.get_object()
+        if not program.check_access(request.user, ACCESS_READ):
+            return HttpResponseForbidden()
+
+        discovery_program = get_discovery_program(program.discovery_uuid)
+        curricula = discovery_program.get('curricula')
+
+        course_runs = []
```
kdmccormick, line 99: Might be good to drop an inline comment here saying something to the effect of "we do not support old-style programs, and we do not go more than one level deep in the program/curricula tree when looking for course runs"

registrar/apps/enrollments/data.py
```diff
@@ -1,65 +1,65 @@
...
+def enroll_in_course(user, course_id, client=None):
     """
     TODO: LMS enrollment API only allows you to enroll
     the currently logged-in user (I think).
     """
     pass
 
 
-def _get_request(url):
+def _get_request(url, client=None):
```
kdmccormick, line 53: With this change we will be re-authenticating with every request. Not saying that's necessarily a bad thing, but I'm wondering why you decided to make this change.

kdmccormick (changes requested):
Good work, just a few comments.

registrar/apps/api/serializers.py
zacharis278: ahh good call I'll translate the output values

registrar/apps/enrollments/data.py
zacharis278: Generally, yes.  We aren't managing a long-term session, this appears to be the model we're using for backend requests elsewhere as well.  That optional client param would allow you to re-use a session if you wanted to make multiple backend requests.

registrar/apps/api/v1/views.py
```diff
@@ -26,6 +28,10 @@ class ProgramReadOnlyViewSet(viewsets.ReadOnlyModelViewSet):
         Return the program associated with the given ``program_key``,
         or 404 if no such program exists.
 
+    /api/v1/programs/{program_key}/courses
+        List all courses associated with the given ``program_key``,
+        or 404 if now such program exists.
```
iloveagent57: s/now/no/

iloveagent57 (commented):
Looks pretty good, I just found one typo, looks like you and Kyle have covered everything else.

registrar/apps/enrollments/data.py
```diff
@@ -1,65 +1,65 @@
...
-
-
-def get_discovery_program(program_uuid):
+def get_discovery_program(program_uuid, client=None):
     """
     Fetches program data from discovery service, given a program UUID.
     """
-    url = urljoin(settings.DISCOVERY_BASE_URL, 'api/v1/programs/{}').format(program_uuid)
-    return _get_request(url)
+    url = urljoin(settings.DISCOVERY_BASE_URL, 'api/v1/programs/{}/').format(program_uuid)
```
kdmccormick, line 30: This will blow up and give a 500 if the program doesn't exist in Discovery. Which is fine... but I think it'd be good to catch the Discovery 404, log an error that Registrar is out of sync with Discovery, and then return a 500.

kdmccormick (changes requested):
One more comment. Looks good otherwise.

registrar/apps/enrollments/data.py
zacharis278: huh, turns out discovery just 500's on this.  Anyway, added a log message in case this happens.

PR #29, opened 2019-03-14
remove org(group) code from enrollments

jansenk:
Remove models and permissions file


kdmccormick (approved):
LGTM :+1: 

I'm realizing now that we could have done this step in the same PR as the previous one. No big deal, but good to note for future model moves/renames.

schenedx (approved):
👍

iloveagent57 (approved):

PR #30, opened 2019-03-14
Remove public_url_root, which is introducing login errors

schenedx:
## Description

According to the conversation in slack channel #hackathon_depr_dop, my change to configuration (https://github.com/edx/configuration/pull/5054) is reverted, and we need this on registrar to make oauth2 work.

### DIFF
--- a/registrar/settings/base.py
+++ b/registrar/settings/base.py
@@ -177,7 +177,6 @@ SOCIAL_AUTH_EDX_OAUTH2_KEY = 'replace-me'
 SOCIAL_AUTH_EDX_OAUTH2_SECRET = 'replace-me'
 SOCIAL_AUTH_EDX_OAUTH2_ENDPOINT = 'replace-me'
 SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT = 'replace-me'
-SOCIAL_AUTH_EDX_OAUTH2_PUBLIC_URL_ROOT = 'replace-me'
 SOCIAL_AUTH_EDX_OAUTH2_LOGOUT_URL = 'replace-me'
 
 # These values are used to make server to server rest api call. Should be fed into edx_rest_api_client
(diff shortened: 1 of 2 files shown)


registrar/settings/devstack.py
```diff
@@ -39,9 +39,7 @@
 SOCIAL_AUTH_EDX_OAUTH2_ISSUER = os.environ.get('SOCIAL_AUTH_EDX_OAUTH2_ISSUER', 'http://localhost:18000')
 SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT = os.environ.get('SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT', 'http://edx.devstack.lms:18000')
 SOCIAL_AUTH_EDX_OAUTH2_LOGOUT_URL = os.environ.get('SOCIAL_AUTH_EDX_OAUTH2_LOGOUT_URL', 'http://localhost:18000/logout')
-SOCIAL_AUTH_EDX_OAUTH2_PUBLIC_URL_ROOT = os.environ.get(
-    'SOCIAL_AUTH_EDX_OAUTH2_PUBLIC_URL_ROOT', 'http://localhost:18000',
-)
+
```
pwnage101: I think you still need to keep this for devstack to work.

pwnage101: otherwise, LGTM

schenedx: @pwnage101 Are you sure that in devstack we need `public_url_root` setting?
I tested the removal on my devstack, and oauth2 still works.
I find it difficult to understand why we need it in devstack, but we don't in stage and prod through configuration.

schenedx: We sync'ed offline and found out that local.py setting file included hostname referencing values, which we should not have. Removed those in this PR.

PR #32, opened 2019-03-14
Use OrganizationGroup-based access-checking on API views

kdmccormick:
[EDUCATOR-4169](https://openedx.atlassian.net/browse/EDUCATOR-4169) @edx/masters-neem 

This PR builds upon the work that installed Django Guardian and created `OrganizationGroup`s. It updates our API endpoints to use this new authorization scheme.

Additionally, it fixes two issues with our current `OrganizationGroup` implementation:
* It fixes the description of `permissions.ORGANIZATION_WRITE_ENROLLMENTS`, which required a migration
* It makes `OrganizationGroups` remove and re-assign all their permissions upon save, instead of simply granting themselves the new permissions. This way, if an `OrganizationGroup` is modified, it does not retain permission assignments from its previous state.

Note: I did not use Guardian's `PermissionRequiredMixin`, as specified in the ticket. It overrides `APIView.dispatch` in a way that breaks authentication.

To-do:
- [x] Implementation
- [x] Tests
- [x] Pylint
- [x] Docstrings
- [ ] ~Swagger UI?~


Makefile
```diff
@@ -103,7 +103,7 @@ coverage: clean
 	$(BROWSER) htmlcov/index.html
 
 test: clean ## run tests and generate coverage report*
-	pytest
+	pytest --ds=registrar.settings.test
```
iloveagent57: Haha, I just ran into this.  You can add this option to the `addopts` section of `pytest.ini`, that way you don't have to worry about remembering to include the option if you, say, run plain `pytest` from your registrar app container.

registrar/apps/api/v1/views.py
```diff
@@ -1,96 +1,200 @@
...
 
-class ProgramReadOnlyViewSet(viewsets.ReadOnlyModelViewSet):
+class AuthMixin(object):
+    """
+    Mixin providing AuthN/AuthZ functionality for all our views to use.
+
+    This mixin overrides `APIView.check_permissions` to use Django Guardian.
+    It replicates, to the extent that we require, the functionality of
+    Django Guardian's `PermissionRequiredMixin`, which unfortunately doesn't
+    play nicely with Django REST Framework.
```
iloveagent57, line 31: What about it doesn't play nicely with DRF?

registrar/apps/core/models.py
```diff
@@ -110,13 +90,29 @@ class Meta(object):
...
-            if self.role is role.name:
-                role.assign_to_group(self, self.organization)
-                break
+        self.refresh_permissions()
+
+    def refresh_permissions(self):
+        """
+        Re-calculates this group's permissions based on its org and role.
+        """
+        for org in Organization.objects.all():
```
iloveagent57: Do you really want to do this for all organizations?

iloveagent57 (commented):
A few general questions for now, looking good so far.

Makefile
kdmccormick: Great solution.

registrar/apps/core/models.py
kdmccormick: I think we have to, unless there's a way to access the Organization that this OrganizationGroup referenced before the save happened.

registrar/apps/api/v1/views.py
kdmccormick: Let's chat tomorrow; easier to explain in person.

registrar/apps/core/models.py
zacharis278: humm, is there a way could alter this so the organization cannot be modified?  I imagine the use case for updating this model is really just to change the role 99% or the time right?

iloveagent57: You might be able to get the set of updated fields for this model using a `post_save` receiver: 
https://docs.djangoproject.com/en/2.1/ref/signals/#post-save

Or you could use the provided `update_fields` kwarg directly in the save method: https://docs.djangoproject.com/en/2.1/ref/models/instances/#django.db.models.Model.save

You might also take a look at [get_objects_for_group](https://django-guardian.readthedocs.io/en/stable/api/guardian.shortcuts.html#get-objects-for-group) if you want to whittle this down to only the set of Organizations that this group has permissions on.

iloveagent57: ...the idea with the `update_fields` being, if the save() didn't modify the organization field of this instance, you don't have to worry about deleting permissions on Organizations that were previously associated with this group.

kdmccormick: @Zacharis278  There's ways to make a model immutable (overriding the `save()` method), but I wasn't able to find any non-hacky ways of making a single field immutable.

kdmccormick:
@iloveagent57 @Zacharis278 I pushed a commit that takes a different approach to refreshing permissions,  avoiding looping through all organizations. Let me know what you think.

registrar/apps/core/models.py
kdmccormick: @iloveagent57 I don't believe `update_fields` always contains all the fields that were updated. If `None` is passed to `save()`, then all fields are updated.

zacharis278:
that change looks good to me 👍

iloveagent57 (approved):
Agreed on that `initial_organization` change, rest looks good 👍

zacharis278 (approved):

PR #34, opened 2019-03-18
Install celery, configure for local usage only.

iloveagent57:
https://openedx.atlassian.net/browse/EDUCATOR-4113

* Installs celery into the project.
* Sets things up for local task invocation only.
* Adds a simple `debug_task`.
* Adds a unit test of the debug_task to make sure that we can appropriately unit test tasks.


registrar/settings/base.py
```diff
@@ -80,6 +80,31 @@
     }
 }
 
+############################# END CELERY #################################3
```
kdmccormick: `### BEGIN CELERY ###`

kdmccormick (approved):
LGTM, pending one nit

kdmccormick:
The ticket has an acceptance criteria "Update settings value to specify Celery broker and queue name". I assume this isn't relevant? Or maybe I missed the part of the code that does this?

iloveagent57:
Good call, I updated the ticket.

PR #33, opened 2019-03-18
Add the PendingOrganizationUserRole model to enable the auto role assignment

schenedx:
## Description

This is the PR to add the PendingUserOrganizationGroup so we can pre-assign user role before the user is created in core.

Also added the django signal to automatically assign newly created user into the proper organization_group if we added data to the PendingUserOrganizationGroup table

@edx/masters-neem Please review

PR #38, opened 2019-03-19
Fix OrganizationGroup adding in admin panel

kdmccormick:
While testing #33 I noticed that I introduced a bug in #32: trying to add an `OrganizationGroup` through the admin panel causes a 500. This is because we set `self._initial_organization = self.organization` in `OrganizationGroup.__init__`, which fails when `self.organization` has not yet been set.

This PR wraps the offending line in a try-catch.


schenedx (approved):
👍

PR #40, opened 2019-03-19
Add missing migration caused by permission prefixing

kdmccormick:


kdmccormick:
Closed in favor of #41

PR #41, opened 2019-03-19
Fix permission naming in Organization model

kdmccormick:


registrar/apps/core/permissions.py
```diff
@@ -7,29 +7,36 @@
 
 APP_PREFIX = 'core.'
 
+# Non-prefixed names of Organization permissions.
+# In general, the prefixed versions (named ORGANIZATION_*_PERM)
```
schenedx: The example should be `ORGANIZATION_APP_PREFIX_PERM` correct?

schenedx (commented):
See my inline question

schenedx (approved):
:+1:

PR #44, opened 2019-03-21
fix segment

jansenk:
I could have sworn that I tested that this worked locally but I must have missed something because I had too many parameters

### DIFF
--- a/registrar/apps/api/segment.py
+++ b/registrar/apps/api/segment.py
@@ -9,16 +9,15 @@ logger = logging.getLogger(__name__)
 
 
 def track(
-    user_id=None,
-    event=None,
+    user_id,
+    event,
     properties=None,
     context=None,
     timestamp=None,
     anonymous_id=None,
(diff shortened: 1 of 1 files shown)


registrar/apps/api/segment.py
```diff
@@ -9,16 +9,15 @@
 
 
 def track(
-    user_id=None,
-    event=None,
+    user_id,
```
jansenk, line 12: also when i was referencing the docs, i saw that these are both required so changed that as well

schenedx (approved):
👍 Change looks good. How do you explain why this works on local, but not in stage nor prod?

iloveagent57 (approved):

jansenk:
> 👍 Change looks good. How do you explain why this works on local, but not in stage nor prod?

I honestly have no idea. The only thing I can think of is that I must have forgotten to re-test the happy path after pulling segment out to the helper class

PR #45, opened 2019-03-21
EDUCATOR-4207 | Update JWT_AUTH setting instead of overwriting.

iloveagent57:
Fixes https://openedx.atlassian.net/browse/EDUCATOR-4207

Thanks to Doug, who noticed that we don't do the same thing that course-discovery does in its `production.py`, which is to ensure that we only update the keys of the `JWT_AUTH` dict, instead of overwriting it completely from the yaml config.

### DIFF
--- a/registrar/settings/production.py
+++ b/registrar/settings/production.py
@@ -12,9 +12,22 @@ ALLOWED_HOSTS = ['*']
 
 LOGGING = get_logger_config()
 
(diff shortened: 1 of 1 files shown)


douglashall (approved):

schenedx (approved):
All other IDA do include this blob of code. We just don't know why. Now we do. 👍

PR #46, opened 2019-03-21
v0 program enrollment

zacharis278:

### DIFF
--- a/registrar/apps/api/serializers.py
+++ b/registrar/apps/api/serializers.py
@@ -43,6 +43,17 @@ class LearnerProgramEnrollmentSerializer(serializers.Serializer):
     status = serializers.CharField()
 
 
+class RequestedLearnerProgramEnrollmentSerializer(serializers.Serializer):
+    """
+    Serializer for request to create a LearnerProgramEnrollment
+    """
+    STATUS_CHOICES = ['pending', 'enrolled']
+
+    student_key = serializers.CharField(allow_blank=False)
+    email = serializers.CharField(allow_blank=False)
+    status = serializers.ChoiceField(allow_blank=False, choices=STATUS_CHOICES)
+
+
 class CourseRunSerializer(serializers.Serializer):
     """
     Serializer for a course run returned from the
--- a/registrar/apps/api/v0/views.py
+++ b/registrar/apps/api/v0/views.py
@@ -7,10 +7,21 @@ from django.core.exceptions import PermissionDenied
 from django.http import Http404
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
 from rest_framework.authentication import SessionAuthentication
-from rest_framework.generics import ListAPIView, RetrieveAPIView
+from rest_framework.exceptions import ValidationError
+from rest_framework.generics import CreateAPIView, ListAPIView, RetrieveAPIView
 from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.status import (
+    HTTP_207_MULTI_STATUS,
+    HTTP_413_REQUEST_ENTITY_TOO_LARGE,
+    HTTP_422_UNPROCESSABLE_ENTITY
+)
 
-from registrar.apps.api.serializers import ProgramSerializer, CourseRunSerializer
+from registrar.apps.api.serializers import (
+    CourseRunSerializer,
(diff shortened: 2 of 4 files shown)


registrar/apps/api/serializers.py
```diff
@@ -43,6 +43,17 @@ class LearnerProgramEnrollmentSerializer(serializers.Serializer):
     status = serializers.CharField()
 
 
+class RequestedLearnerProgramEnrollmentSerializer(serializers.Serializer):
```
schenedx, line 46: Can we have this serializer to be in a different file than the serializer used for v1 versions? Or better, just write a new serializer file in the v0 folder? I would like them to be separate

schenedx (approved):
Besides the minor serializer suggestion, rest LGTM

registrar/apps/api/serializers.py
zacharis278: Humm so do we want to duplicate every serializer or just on an as-needed basis?  This isn't the only one we use, I'd be worried we're unnecessarily repeating ourselves.

Or maybe just do it for input serializers like this and leave the models as-is?

schenedx: Ah, I didn't realize this serializer could be used for V1 as well. In that case, you can ignore my comment above

registrar/apps/api/v0/views.py
```diff
@@ -114,3 +125,66 @@ def get_queryset(self):
...
+     * 413: Payload too large, over 25 students supplied.
+     * 422: Invalid request, unable to enroll students.
+    """
+    authentication_classes = (JwtAuthentication, SessionAuthentication)
+    permission_classes = (IsAuthenticated,)
+    serializer_class = RequestedLearnerProgramEnrollmentSerializer
+
+    def post(self, request, *args, **kwargs):
+        """ Enroll up to 25 students in program """
+        if not self.program.managing_organization.metadata_readable:
```
kdmccormick, line 151: Should be checking against `.enrollments_writeable` here.

registrar/apps/api/v0/views.py
```diff
@@ -114,3 +125,66 @@ def get_queryset(self):
...
+
+        response = {}
+        error_status = None
+
+        if not isinstance(request.data, list):
+            raise ValidationError()
+
+        if len(request.data) > 25:
+            return Response(
+                'enrollement limit 25', HTTP_413_REQUEST_ENTITY_TOO_LARGE
```
kdmccormick, line 162: `enrollment` (sp.)

registrar/apps/api/v0/views.py
```diff
@@ -114,3 +125,66 @@ def get_queryset(self):
...
+    permission_classes = (IsAuthenticated,)
+    serializer_class = RequestedLearnerProgramEnrollmentSerializer
+
+    def post(self, request, *args, **kwargs):
+        """ Enroll up to 25 students in program """
+        if not self.program.managing_organization.metadata_readable:
+            raise PermissionDenied()
+
+        response = {}
+        error_status = None
```
kdmccormick, line 155: To get into the weeds of the API spec for a bit: 
* `error_status` should be `422` if no students are successfully recorded. 
* `error_status` should be `207` if some are successfully recorded
* `error_status` should be `200` if some all are successfully recorded

In this implementation, `error_status` will be `207` if all of the students are given `invalid-status`, whereas the spec says it should be `422`.

registrar/apps/api/serializers.py
kdmccormick, line 46: I suggest `ProgramEnrollmentRequestSerializer`. Rolls off the tongue a little easier.

registrar/apps/api/v0/views.py
```diff
@@ -114,3 +125,66 @@ def get_queryset(self):
...
+                    response[student_key] = 'duplicated'
+                    error_status = HTTP_207_MULTI_STATUS
+                else:
+                    response[student_key] = enrollee['status']
+            else:
+                try:
+                    if 'status' in enrollee_serializer.errors:
+                        response[enrollee['student_key']] = 'invalid-status'
+                    else:
+                        response[enrollee['student_key']] = 'internal-error'
```
kdmccormick, line 183: I think missing `email` should raise a 422, just like missing `student_key` would. `'internal-error'` is for unexpected processing errors on our side.

PR #48, opened 2019-03-21
Strip out local Docker development & other cleanup

kdmccormick:
@edx/masters-neem 

Now that Registrar runs seamlessly on Devstack, one could argue that we do not need to support a local, standalone Docker development environment for Registrar. If one wants to run Registrar locally, they can follow the instructions in the `README` to run it within a virtualenv.

Removing `docker-compose` from our requirements removes a dependency on PyYAML 3, allowing us to pin PyYAML at 5.1 without breaking `make upgrade`.


iloveagent57 (commented):
Looks great!  I think we may want to update this line: https://github.com/edx/registrar/blob/a4f841c0163daddd512538b5b00f2bb4fcff9fa4/manage.py#L11 to point to `registrar.settings.devstack` instead of the local settings, right?

kdmccormick:
@iloveagent57 Agreed. Ready for re-review.

README.rst
```diff
@@ -64,6 +33,57 @@ To view all commmands::
...
+
+Navigate in, create a Python 3 virtualenv, and activate it::
+
+  cd registrar
+  virtualenv --python=python3 venv
+  source venv/bin/activate
+
+Ensure local settings are used instead of Devstack settings::
+
+  export DJANGO_SETTINGS_MODULE=registrar.settings.local
```
iloveagent57, line 66: Great job thinking about this.

iloveagent57 (approved):
Nice job 👍

README.rst
schenedx: @kdmccormick You don't need to address here, but I believe in the local settings, we do not have value specified for registrar to connect to its dependent services like LMS or Course discovery. That needs to be explained to a person who looks to do local setup of the service. I mention this so you are aware

kdmccormick: @schenedx You are correct. I mention that in the `README`. My thought is we don't need to worry about it as long as we let potential stand-alone users know.

kdmccormick: https://github.com/edx/registrar/pull/48/files#diff-88b99bb28683bd5b7e3a204826ead112R39

PR #49, opened 2019-03-22
EDUCATOR-4175 | Program enrollment v0 GET resource.

iloveagent57:
https://openedx.atlassian.net/browse/EDUCATOR-4175

* Introduces a `FakeTask` class.
* Associates fake task ids with any fake program with readable enrollments data.
* Exposes GET method in the `MockProgramEnrollmentView`

### DIFF
--- a/registrar/apps/api/v0/data.py
+++ b/registrar/apps/api/v0/data.py
@@ -28,6 +28,11 @@ FakeCourseRun = namedtuple('FakeCourseRun', [
     'marketing_url',
 ])
 
+FakeTask = namedtuple('FakeTask', [
+    'job_id',
+    'job_url'
+])
+
 
 def _program(org, program_key, title):
     """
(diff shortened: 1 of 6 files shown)


registrar/apps/api/v0/views.py
```diff
@@ -195,3 +224,26 @@ def post(self, request, *args, **kwargs):
             return Response(response, HTTP_207_MULTI_STATUS)
         else:
             return Response(response)
+
+    def get_object(self):
```
kdmccormick: Would be preferable to have this rolled into `retrieve`.

registrar/apps/api/v0/data.py
```diff
@@ -153,3 +158,14 @@ def _course_run(title, org, course, run="Spring2050"):
     for program_key, course_run_keys
     in _FAKE_PROGRAM_COURSE_RUN_KEY_BODIES.items()
 }
+
+
+def _fake_job_id(program_key):
+    return 'fake-job-for-{}'.format(program_key)
+
+
+FAKE_TASK_IDS_BY_PROGRAM = {
```
kdmccormick, line 167: I think a dict referring to static (but real) UUIDs would be a more faithful mock

iloveagent57: Yeah, and actually we'll need to do that if we want to use the real Serializer class, because that thing will eventually want to use a `UUIDField`, not just a `CharField`.

kdmccormick: Yep. Also I recommend the name `JobSerializer` -- we'll want to keep "jobs" and "tasks" conceptually distinct.

kdmccormick (approved):
Great fake UUIDs. Taste just the same as the real ones, can't tell the difference. :+1:

PR #50, opened 2019-03-22
 Fix API doc permission-checking

kdmccormick:
Swagger API docs broke when I implemented permissions. This PR fixes that.

Also, it makes `/login` by default redirect to `/api-docs/` instead of `/admin`, because most users who go straight to `/login` would rather see documentation than a 403.

@edx/masters-neem

### DIFF
--- a/registrar/apps/api/v1/views.py
+++ b/registrar/apps/api/v1/views.py
@@ -7,7 +7,9 @@ import logging
 from django.core.exceptions import ImproperlyConfigured, PermissionDenied
 from django.http import Http404
 from django.shortcuts import get_object_or_404
+from django.urls import resolve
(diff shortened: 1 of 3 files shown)


registrar/apps/api/v1/views.py
```diff
@@ -74,6 +80,23 @@ def check_permissions(self, request):
         else:
             raise PermissionDenied()
 
+    def check_doc_permissions(self, request):
+        """
+        Check whether the endpoint being requested should show up in the
```
iloveagent57, line 85: Nice job with this docstring.

iloveagent57 (approved):

schenedx (approved):
👍

PR #53, opened 2019-03-26
v0 post for course enrollments

jansenk:
pulled validate_and_echo_statuses out into it's own class so it can be used between multiple views

### DIFF
--- a/registrar/apps/api/v0/views.py
+++ b/registrar/apps/api/v0/views.py
@@ -25,6 +25,7 @@ from registrar.apps.api.serializers import (
     ProgramSerializer,
     ProgramEnrollmentRequestSerializer,
     ProgramEnrollmentModificationRequestSerializer,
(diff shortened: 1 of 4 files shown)


registrar/apps/api/v0/views.py
```diff
@@ -133,7 +152,59 @@ def get_queryset(self):
...
+
+        if not isinstance(request.data, list):
+            raise ValidationError()
+
+        if len(request.data) > 25:
+            return Response(
+                'enrollment limit 25', HTTP_413_REQUEST_ENTITY_TOO_LARGE
+            )
+
+        response = {}
```
kdmccormick: NIT: `results` might be preferable here, to disambiguate it from the `Response` that it gets wrapped in.

kdmccormick (approved):
Looks good! Just one optional nit.

jansenk:
added get_serializer_class to the views

PR #54, opened 2019-03-26
patch course enrollments

jansenk:
pointing at my previous branch for the sake of the diff

added an optional dict argument to validate_and_echo_statuses to allow us to return enrolled-processing as a status since enrolled is not a valid status to return from this endpoint

### DIFF
--- a/registrar/apps/api/v0/views.py
+++ b/registrar/apps/api/v0/views.py
@@ -26,6 +26,7 @@ from registrar.apps.api.serializers import (
     ProgramEnrollmentRequestSerializer,
     ProgramEnrollmentModificationRequestSerializer,
     CourseEnrollmentRequestSerializer,
+    CourseEnrollmentModificationRequestSerializer,
 )
 from registrar.apps.api.v0.data import (
     FAKE_ORG_DICT,
@@ -153,7 +154,10 @@ class MockProgramCourseListView(MockProgramSpecificViewMixin, ListAPIView):
 
 
 class EchoStatusesMixin(object):
-    """ Provides the validate_and_echo_statuses function """
+    """
+    Provides the validate_and_echo_statuses function
+    Classes that inherit from EchoStatusesMixin must implement get_serializer_class
+    """
 
     def validate_and_echo_statuses(self, request):
         """ Enroll up to 25 students in program/course """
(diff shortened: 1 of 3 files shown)


registrar/apps/api/v0/views.py
```diff
@@ -304,7 +308,14 @@ class MockCourseEnrollmentView(APIView, MockProgramCourseSpecificViewMixin, Echo
...
+        elif self.request.method == 'PATCH':
+            return CourseEnrollmentModificationRequestSerializer
 
     def post(self, request, *args, **kwargs):
         self.course  # trigger 404  # pylint: disable=pointless-statement
         return self.validate_and_echo_statuses(request)
+
+    def patch(self, request, *args, **kwargs):  # pylint: disable=unused-argument
+        self.course  # trigger 404  # pylint: disable=pointless-statement
+        status_replacements = {'enrolled': 'enrolled-processing'}
```
kdmccormick: `enrolled-waiting`, per the spec

(although that's subject to change soon)

registrar/apps/api/v0/views.py
```diff
@@ -155,7 +156,7 @@ def get_queryset(self):
 class EchoStatusesMixin(object):
     """ Provides the validate_and_echo_statuses function """
 
-    def validate_and_echo_statuses(self, request):
+    def validate_and_echo_statuses(self, request, status_replacements=None):
```
kdmccormick: `status_replacements={}` would remove the requirement of checking `if status_replacements`

registrar/apps/api/v0/views.py
```diff
@@ -155,7 +156,7 @@ def get_queryset(self):
 class EchoStatusesMixin(object):
     """ Provides the validate_and_echo_statuses function """
```
kdmccormick: Can you add a note here that says that subclasses of `EchoStatusesMixin` must implement `get_serializer_class`?

registrar/apps/api/v0/views.py
```diff
@@ -304,7 +308,14 @@ class MockCourseEnrollmentView(APIView, MockProgramCourseSpecificViewMixin, Echo
     def get_serializer_class(self):
```
kdmccormick, line 308: ~Something I realized when working on my own PR: If we make the enrollment views subclass `GenericAPIView` instead of `APIView`, Swagger will call the `get_serializer_class` and the JSON input field in the Swagger UI will start working again.~

Actually, it looks like my suggestion would bring back the browser-only `AttributeError` we were seeing before. I'm gonna try to figure out a way to make the browsable API play nice with our API, but in the mean time, don't worry about making this change.

kdmccormick (changes requested):

registrar/apps/api/v0/views.py
jansenk: That’s causes a pylint warning and I don’t think this little block is too bad

registrar/apps/api/v0/views.py
jansenk: enrolled-processing is in the spec too:
Screen Shot 2019-03-27 at 10 03 06 AM

registrar/apps/api/v0/views.py
kdmccormick: Gotcha, no biggie

registrar/apps/api/v0/views.py
kdmccormick: Ah. It was removed from POST at Simon's request but I must have forgotten to remove it from PATCH. My b. Sorry about all these discrepancies in the spec.

jansenk: as per conversation i removed this stuff and am just echoing 'enrolled' again

schenedx (approved):
👍

PR #56, opened 2019-03-26
Implement V0 fake asynchronous jobs

kdmccormick:
# Real Fake Jobs
real fake door salesman

[EDUCATOR-4179](https://openedx.atlassian.net/browse/EDUCATOR-4179)
@edx/masters-neem 

### This PR:
* Creates infrastructure for fake jobs using Django cache
* Implements job retrieval (`/api/v0/jobs/{job_id}`)
* Updates program enrollment GET to use fake jobs
* Adds some fake, static JSON enrollment files to serve as program enrollment GET results
* Does some minor refactoring of existing tests

### On result file expiration
Right now, job statuses age-off from the cache after 24 hours (a number which is easy to change), but the result JSON files are static and always available. Implementing dynamic result files with age-off would require more work than I felt was warranted for the V0 API.

### Enrollment data script

I added a `scripts/` directory and put in it the Python script I used to generate the JSON files. I can remove it if you guys don't think it belongs in the repository.


registrar/apps/api/tests/mixins.py
```diff
@@ -51,3 +52,61 @@ def generate_jwt_header(self, user, admin=False, ttl=5, **overrides):
...
+        """
+        kwargs = {'follow': True}
+        if user:
+            kwargs['HTTP_AUTHORIZATION'] = self.generate_jwt_header(
+                user, admin=user.is_staff,
+            )
+        if data:
+            kwargs['data'] = json.dumps(data)
+            kwargs['content_type'] = 'application/json'
+        full_path = (
```
iloveagent57: you can do this in the same number of lines with a regular `if ... else ...` block :)

scripts/fake_program_enrollments.py
```diff
@@ -0,0 +1,108 @@
...
+        email_domain (str): the part of the email after the @
+
+    Returns: str
+    """
+    return '{}{}@{}'.format(
+        fake.first_name()[0], fake.last_name(), email_domain,
+    ).lower()
+
+
+exit(main(sys.argv[1:]))
```
iloveagent57: Is this a python 3 convention (as opposed to `if __name__ == '__main__'`)?

iloveagent57 (approved):
This generally looks good - I know you're still committing some minor cleanup and fixes.

scripts/fake_program_enrollments.py
kdmccormick: Nope, it's not a convention. I call `exit` with a return code for BASH interoperability. 

Wrapping it in `if __name__ == '__main__'` is a good idea, too.

schenedx (commented):
The code LGTM
As the result of our offline conversation, please move all the static files under the top level `static` folder.

schenedx (approved):
👍

PR #59, opened 2019-04-03
WIP some progress towards implementing V1 program enrollment GET

kdmccormick:
[EDUCATOR-3987](https://openedx.atlassian.net/browse/EDUCATOR-3987)

This is a work-in-progress implementation of the V1 program enrollment GET endpoint (`GET /api/v1/programs/{program_key}/enrollments`). It's not ready for review or merging.

I'm putting this work down for now because:
* It depends on the design of the LMS program enrollment API
* It depends on the file storage infrastructure (S3 for prod, file system for local)
* I've found that `django-user-tasks` is not compatible with our job status API design (which includes getting `original_url`) unless we create a new model, at which point we probably shouldn't even bother using `django-user-tasks`.


kdmccormick:
Closed in favor of #66

PR #60, opened 2019-04-05
add yaml based openapi spec

zacharis278:
### [EDUCATOR-4227](https://openedx.atlassian.net/browse/EDUCATOR-4227)

1. Replaces django-rest-swagger with a standalone version of swagger-ui
2. swagger page now driven by yaml specification instead of DRF's automatically generated output
    a. request body / params now function on all endpoints
    b. split out v1 and mock v0 routes
3. api spec includes annotations for integration with AWS API-Gateway
4. added some linting/validation tasks for yaml files


openedx.yaml
```diff
@@ -1,8 +1,8 @@
 # This file describes this Open edX repo, as described in OEP-2:
 # http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
-
+---
```
schenedx: Why do we need these 3 `-`?

schenedx (changes requested):
Can you add instructions to the READ.md file on how to generate and maintain the swagger api yaml file?

openedx.yaml
zacharis278: Was thinking yaml needed to indicate the document beginning with that since the linter was yelling at me.  Looks like it's not strictly required in this case, and after poking around we aren't doing it elsewhere.  I'll just remove it along with the linting rule for the sake of consistency

nedbat: @Zacharis278 I'm curious what linter complained about this?

registrar/api_renderer.py
```diff
@@ -0,0 +1,32 @@
...
+from django.conf import settings
+from django.shortcuts import render
+
+
+module_dir = os.path.dirname(__file__)
+spec_file = open(os.path.join(module_dir, '../api.yaml'))
+API_SPEC = yaml.safe_load(spec_file.read())
+
+
+# pylint: disable=missing-docstring
```
iloveagent57: Nit: just give it a short docstring?

iloveagent57 (approved):
This looks good to me, after you add some stuff to the README like Simon suggested.

openedx.yaml
zacharis278: @nedbat I added yamllint to this project to validate the swagger spec.  The default ruleset warns `missing document start "---"  (document-start)`

zacharis278:
@schenedx I've updated the readme with some additional information on this

schenedx (approved):
👍 
Thank you.

PR #61, opened 2019-04-09
Flesh out YAML API documentation

kdmccormick:
@edx/masters-neem 

**How to test**
* `git checkout kdmccormick/api-docs`
* `docker restart edx.devstack.registrar`
* Navigate to http://localhost:18734/login, log in as edx@example.com so all endpoints are visible
* Navigate to http://localhost:18734/api-docs and make sure the look right and are easy to understand

### DIFF
--- a/.api-generated.yaml
+++ b/.api-generated.yaml
@@ -19,23 +19,132 @@ program_path_param:
   name: program_key
   required: true
   type: string
+  description: edX program key
+  example: uexample-master-of-science
 course_path_param:
   in: path
   name: course_id
   required: true
   type: string
-course_enrollment_body:
+  description: edX course run identifier
+  example: course-v1:ExU+Science-101+Fall2050
+enrollment_body:
   in: body
   name: body
   required: true
   schema:
-    $ref: '#/models/CourseEnrollmentList'
-program_enrollment_body:
-  in: body
-  name: body
-  required: true
-  schema:
-    $ref: '#/models/ProgramEnrollmentList'
+    $ref: '#/models/EnrollmentRequests'
+
+post_enrollment_responses:
+  200:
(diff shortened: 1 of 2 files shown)


.api-generated.yaml
```diff
@@ -19,23 +19,132 @@ program_path_param:
...
+      $ref: '#/models/EnrollmentResult'
+    examples:
+      application/json:
+        student_0128fe4a: invalid-status
+        student_aae45c81: not-found
+
+student_key:
+  type: string
+  description: >
+    key that uniquely identifies student within organization.
```
reillz10: Need a capital `k` to start the sentence

.api-generated.yaml
```diff
@@ -537,54 +1015,109 @@ models:
...
     type: array
     items:
       type: object
       properties:
         student_key:
           type: string
-        email:
-          type: string
+          description: >
+            key that uniquely identifies student within organization.
```
reillz10: Same

api.yaml
```diff
@@ -20,23 +20,132 @@ program_path_param: &program_path_param
...
+      $ref: '#/models/EnrollmentResult'
+    examples:
+      application/json:
+        student_0128fe4a: invalid-status
+        student_aae45c81: not-found
+
+student_key: &student_key
+  type: string
+  description: >
+    key that uniquely identifies student within organization.
```
reillz10: same

reillz10 (changes requested):
From what I can see, just a couple of "k"'s that need to be capitalized. Other than that looks good!

api.yaml
kdmccormick: Good call; ready for another look!

schenedx (approved):
👍

PR #62, opened 2019-04-10
Include external service base url variable names in base settings.

iloveagent57:
## Description
https://openedx.atlassian.net/browse/EDUCATOR-4232

This change makes it really clear that `LMS_BASE_URL` and `DISCOVERY_BASE_URL` are required configuration variables in all environments (the changes from the configuration PRs linked in the ticket will cause the stage/prod environments to have the values that we want).


schenedx (approved):
👍

PR #65, opened 2019-04-11
Allow global staff to access all programs in mock API

iloveagent57:
## Description

Affects only the v0/mock API: A user granted the `ORGANIZATION_READ_METADATA` permission (doing it through a group, here) can list and retrieve metadata for all programs (if excluding `org` query param), even for U. of Perezburgh.


zacharis278 (approved):
very useful, thanks for adding this

PR #66, opened 2019-04-16
Implement V1 program enrollment GET

kdmccormick:
### About
Team: @edx/masters-neem 
Issues: [EDUCATOR-3987](https://openedx.atlassian.net/browse/EDUCATOR-3987) and [EDUCATOR-4111](https://openedx.atlassian.net/browse/EDUCATOR-4111).
Necessary configuration changes: [Configuration PR](https://github.com/edx/configuration/pull/5095) and [Internal PR](https://github.com/edx/edx-internal/pull/841)

This PR does a few things:
* Adds a V1 endpoint for getting program enrollments in a JSON/CSV file
* Adds a "job" system on top of django-user-tasks
* Adds a V1 endpoint for retrieving job statuses
* Uses Boto3 to upload files to Amazon S3 and provide signed URLs for limited-time access
* Uses the Moto library to mock out Boto3 in unit tests
* Uses basic FileSystemStorage for storing generated files when developing locally

### An open issue

For local development, I'm not sure how to return an absolute URL to a job-generated file. I'm thinking that we don't worry about this for now, and revisit it when we address the greater issue of having a configured `BASE_URL`.

### Manually testing
First, create a program in Registrar's Django admin.

Then, mock out the LMS get enrollments API by adding an early return to `_get_all_paginated_results` in registrar/apps/enrollments/data.py:
```python
def _get_all_paginated_results(url, client=None):
    """
    Builds a list of all results from a cursor-paginated endpoint.

    Repeatedly performs request on 'next' URL until 'next' is null.
    """
    return [
        {'student_key': 'a', 'status': 'enrolled', 'account_exists': True},
        {'student_key': 'b', 'status': 'pending', 'account_exists': False},
        {'student_key': 'c', 'status': 'canceled', 'account_exists': True},
    ]
    # ... the rest of the function
```

Next, hit http://localhost:18734/api/v1/programs/{program_key}/enrollments/ with the key of the program you created.

It should return 202 and a job ID and URL. Hit the job URL. It should return a Succeeded status and a result URL.

The result URL is actually a path relative to the host. So, append the path (e.g., `/media/job-results/abcd...json`) to `http://localhost:18734/`, which should give you the JSON file.

Finally, repeat the process with `?fmt=json` and `?fmt=csv` appended to the enrollments URL.


registrar/apps/api/serializers.py
```diff
@@ -15,7 +19,7 @@ class ProgramSerializer(serializers.ModelSerializer):
     """
     Serializer for Programs.
     """
-    program_key = serializers.CharField(source='key')
+    program_key = serializers.SlugField(source='key')
```
schenedx: Why we use Slugfield here instead of the charfield we have specified in Model?

registrar/apps/api/serializers.py
```diff
@@ -24,64 +28,36 @@ class Meta:
...
-    student_key = serializers.CharField(allow_blank=False)
-    status = serializers.ChoiceField(allow_blank=False, choices=STATUS_CHOICES)
+    student_key = serializers.CharField()
+    status = serializers.ChoiceField(choices=PROGRAM_ENROLLMENT_STATUSES)
 
 
 class ProgramEnrollmentModificationRequestSerializer(serializers.Serializer):
     """
-    Serializer for request to modify a LearnerProgramEnrollment
+    Serializer for request to modify a program enrollment in LMS
```
schenedx: #NIT I don't think it's necessary to say `in LMS`. What if that changes in the future somehow?

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -300,3 +311,122 @@ def test_get_program_with_multiple_courses(self):
...
+        self.assertEqual(response.status_code, 202)
+        job_response = self.get(response.data['job_url'], self.hum_admin)
+        self.assertEqual(job_response.status_code, 200)
+        self.assertEqual(job_response.data['state'], 'Succeeded')
+        self.assertTrue(job_response.data['result'].endswith('.json'))
+
+    @_get_enrollments_mock
+    @_SAVE_FILE_MOCK
+    @ddt.data('json')  # , 'csv') TODO test CSVs
+    def test_formats(self, fmt, _mock_get_program_enrollments):
```
schenedx: #NIT, instead of `fmt`, it is clearer to just spell it out as `format` or `file_format`

registrar/apps/core/jobs.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    TODO docstring
+    """
+    artifacts = UserTaskArtifact.objects.filter(
+        status=task_status, name=_RESULT_ARTIFACT_NAME
+    )
+    if artifacts:
+        if artifacts.count() > 1:
+            logger.error(
+                'Multiple UserTaskArtifacts for job ' +
```
schenedx, line 104: Can you also put in the message saying something like `We return the first artifact of the job`?
It will be good to log the first artifact URL as well for debugging purpose.

registrar/apps/core/jobs.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+    task_status.succeed()
+
+
+def post_job_failure(job_id, message):
+    """
+    TODO docstring
+    """
+    task_status = UserTaskStatus.objects.get(task_id=job_id)
+    _assert_in_progress(job_id, task_status)
```
schenedx: If we don't catch this exception, what would be the api behavior? Do we throw 500 back to user?

registrar/apps/enrollments/data.py
```diff
@@ -61,6 +107,6 @@ def _get_request(url, client=None):
     response = client.get(url)
 
     if response.status_code == 200:
-        return response.json()
+        return response
```
schenedx: I agree with this. We should make the helper function here do one thing only.

schenedx (commented):
See my inline comments. The logic looks mostly good.
1. Please fill out all the TODOs
2. What is your plan to manually test? How to ensure this works on both local machine, and when it deploys to stage, it would work there?
3. What if the request on LMS return 404 or unsuccessful, what would the job status be, failed?

registrar/apps/api/serializers.py
kdmccormick: Haha, I think it was wishful thinking that we used a `SlugField` on the model. I'll change this back.

registrar/apps/api/serializers.py
kdmccormick: :+1:

registrar/apps/core/jobs.py
kdmccormick: ~Yes. If the task isn't in progress, that would be an anomalous condition that would warrant a 500, in my opinion.~ 

Correction: This function is called within the task, so the user wouldn't see a 500. Instead, they would see that their job had failed.

schenedx: Yeah, that make sense. Sounds good.

kdmccormick:
@schenedx Thanks for the pre-review!

1. Will do.
2. This is a little tricky. Locally, I've just been mocking the LMS API by adding `return <a bunch of JSON>` to `get_program_enrollments`, which isn't great. On stage/prod, I'm not sure how this will work at all without the LMS API. Any suggestions here would be appreciated.
3. Yep, in `list_program_enrollments` we catch all HTTPErrors from the LMS API, log them, and mark the job as Failed.

.travis.yml
```diff
@@ -7,6 +7,7 @@ cache: pip
 before_install:
     - "export DISPLAY=:99.0"
     - "sh -e /etc/init.d/xvfb start"
+    - "sudo rm -f /etc/boto.cfg"
```
schenedx: Why should we delete the `boto.cfg`? What error does this cause?

api.yaml
```diff
@@ -45,6 +45,14 @@ course_path_param: &course_path_param
   type: string
   description: edX course run identifier
   example: 'course-v1:ExU+Science-101+Fall2050'
+format_param: &format_param
+  in: query
+  name: fmt
```
schenedx, line 50: #NIT why not just use full `format` word for the query string parameter?

registrar/apps/api/serializers.py
```diff
@@ -115,5 +91,5 @@ class JobStatusSerializer(serializers.Serializer):
     }
 
     created = serializers.DateTimeField()
-    state = serializers.ChoiceField(allow_blank=False, choices=STATUS_CHOICES)
-    result = serializers.URLField(allow_null=True)
+    state = serializers.ChoiceField(choices=STATUS_CHOICES)
+    result = serializers.URLField()
```
schenedx: Why is the result field no longer nullable?

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -1,90 +1,89 @@
...
-class RegistrarAPITestCase(APITestCase, RequestMixin):
+class RegistrarAPITestCase(APITestCase):
     """ Base for tests of the Registrar API """
 
     api_root = '/api/v1/'
 
-    def setUp(self):
-        super(RegistrarAPITestCase, self).setUp()
+    @classmethod
+    def setUpClass(cls):
```
schenedx, line 43: Just curious, why did you decide to change the method `setUp` to `setUpClass`?

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -96,10 +95,35 @@ def mock_api_response(self, url, response_data, method='GET', response_code=200)
...
+    Mixin for classes that need to access S3 resources.
+
+    Enables S3 mock and creates default bucket before tests.
+    Disables S3 mock afterwards.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls._s3_mock = moto.mock_s3()
```
schenedx, line 109: Nice~

registrar/apps/api/v1/views.py
```diff
@@ -241,12 +318,36 @@ class ProgramEnrollmentView(ProgramSpecificViewMixin, APIView):
...
+            return perms.ORGANIZATION_READ_ENROLLMENTS
         if request.method == 'POST':
             return perms.ORGANIZATION_WRITE_ENROLLMENTS
+        return []
+
+    def get(self, request, *args, **kwargs):
+        """
+        Submit a user task that retrieves program enrollment data.
+        """
+        file_format = request.query_params.get('fmt', 'json')
```
schenedx, line 337: #NIT why not just use `format`?

registrar/apps/core/job_results.py
```diff
@@ -0,0 +1,97 @@
+""" Utilities for storing the results of jobs. """
```
schenedx: The name of this file does not match the class names in this file. Maybe rename the file to `job_storage.py`?

registrar/apps/core/jobs.py
```diff
@@ -0,0 +1,163 @@
...
+        message: the reason the job failed
+    """
+    task_status = UserTaskStatus.objects.get(task_id=job_id)
+    _assert_in_progress(job_id, task_status)
+    log_message = "Job {} failed. {}".format(job_id, message)
+    logger.error(log_message)
+    task_status.fail(log_message)  # Creates UserTaskArtifact with name "Error"
+
+
+def _assert_in_progress(job_id, task_status):
```
schenedx: #NIT I just don't like the word `assert` when it's not related to testing.
How about `job_should_in_progress`

registrar/settings/test.py
```diff
@@ -29,9 +29,16 @@
...
-CELERY_TASK_ALWAYS_EAGER = True
+CELERY_ALWAYS_EAGER = True
 
 # Results
 CELERY_IGNORE_RESULT = True
 
 # END CELERY
+
+# Media
+DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
```
schenedx, line 40: Is this needed for `moto` library?

registrar/apps/core/jobs.py
```diff
@@ -0,0 +1,163 @@
...
+            `self` (the Celery task instance), `job_id`, and `user_id`.
+            They will be passed in automatically. The function may take
+            additional parameters.
+        *args/**kwargs: Additional arguments to pass to task_fn.
+
+    Returns: str
+        UUID-4 job ID string.
+    """
+    job_id = str(uuid.uuid4())
+    task_fn.apply_async([job_id, user.id] + list(args), kwargs, task_id=job_id)
```
schenedx, line 52: Should we think about retry on the job? Or we should off load that to the client?

schenedx (changes requested):
I think the majority of this is good to go. 
See my inline comments.
I am also totally fine with making the get endpoint NOT resilient against LMS failures. If we have need to make it resilient, we will do so later.
Great work!

registrar/apps/api/v1/views.py
kdmccormick: Django Rest Framework automatically interprets the `format` parameter in a way inconsistent with what we want to do: https://www.django-rest-framework.org/api-guide/format-suffixes/

We could disable it, allowing us to use `format` instead of `fmt`, but that would disable the DRF `format` parameter across all of Registrar.

Do you have thoughts on whether we should stick with `fmt` or disable DRF's `format` for all Registrar APIs?

kdmccormick: Note: DRF also allows us to _rename_ their builtin `format` parameter.

schenedx: I didn't know that. Knowing what you said, I am fine letting this staying at `fmt`

registrar/apps/api/v1/tests/test_views.py
kdmccormick: Good question. `setUp` is run once per test, and `seUpClass` is run once per class. None of the records created in `setUp` were being modified within any of the tests, so it made sense to create them once per class instead of once per test . The change sped up test execution noticeably on my computer.

api.yaml
kdmccormick: (See other comment)

.travis.yml
kdmccormick: When I first pushed this code I got a weird error on Travis CI. This line is a patch, for what seems to be a currently unresolved issue: https://github.com/travis-ci/travis-ci/issues/7940

kdmccormick: You know, I'll add a comment to that line.

registrar/apps/api/serializers.py
kdmccormick: That's a mistake. Good catch.

registrar/settings/test.py
kdmccormick: Yup. We need our code to _try_ to use S3, so that Moto is able to mock out those calls. If we use FileSystemStorage, then the Moto mock won't do anything.

registrar/apps/core/jobs.py
kdmccormick: Makes sense. How about `_affirm_job_in_progress`? I like to start method names with verbs.

registrar/apps/core/jobs.py
kdmccormick: Hm, the only job failure scenario I see that retrying may fix would be if the LMS was temporarily unresponsive. Is that something that we expect to happen often enough for it to bother partners? If so, we can automatically retry. If not, I think one-try is fine for now (knowing we can change it later).

registrar/apps/api/v1/tests/test_views.py
schenedx: Ah. Got ya. Cool. I'll keep this in mind.

registrar/apps/core/jobs.py
schenedx: Yes. That name sounds good

registrar/apps/core/jobs.py
schenedx: One try is fine now. Let's observe and change later if necessary.

schenedx (approved):
👍

PR #67, opened 2019-04-16
v1_mock

jansenk:
change all uses of `v0` to `v1_mock`, except for when used in urls, in which case `v1-mock` (`-` vs `_`) is used instead


zacharis278 (approved):
👍

kdmccormick (approved):
Good stuff :+1:

PR #72, opened 2019-04-23
Use a new yaml_merge.py tool to properly generate the expanded api.yaml file

schenedx:
## Description

With this change, we can use the new yml_merge.py utility to expand the api.yaml. The tool would leverage both the pyyaml and ruamel.ymal "merge-expand" option to get the job done. However, the output won't have any comments because pyyaml strip out comments in api.yaml.

@edx/masters-neem Please review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,8 +69,8 @@ test: clean ## run tests and generate coverage report
 	pytest
 
 quality: ## run Pycodestyle and Pylint
-	pycodestyle registrar *.py
-	pylint --rcfile=pylintrc registrar *.py
+	pycodestyle registrar scripts
+	pylint --rcfile=pylintrc registrar scripts
 	yamllint *.yaml
 
 pii_check: ## check for PII annotations on all Django models
@@ -112,7 +112,7 @@ detect_changed_source_translations: ## check if translation files are up-to-date
 validate_translations: fake_translations detect_changed_source_translations ## install fake translations and check if translation files are up-to-date
 
 api_generated: ## generates an expanded verison of api.yaml for consuming tools that cannot read yaml anchors
-	yaml merge-expand api.yaml .api-generated.yaml
+	python scripts/yaml_merge.py api.yaml .api-generated.yaml
 
 validate_api_committed: ## check to make sure any api.yaml changes have been committed to the expanded document
-	bash -c "diff .api-generated.yaml <(yaml merge-expand api.yaml -)"
+	bash -c "diff .api-generated.yaml <(python scripts/yaml_merge.py api.yaml -)"
--- a/.api-generated.yaml
+++ b/.api-generated.yaml
@@ -1,81 +1,18 @@
-################################################################################
-# Please read before updating!                                                 #
-#   api.yaml -                                                                 #
-#       This is the source of the api spec.  Any api-doc changes should be     #
-#       made to this file      .                                               #
-#   .api-generated.yaml -                                                      #
-#       This is an expanded version of api.yaml since the consuming tools are  #
-#       unable to parse YAML anchors. This file is generated using:            #
-#         `make api_generated`                                                 #
-################################################################################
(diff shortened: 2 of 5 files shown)


yml_merge.py
```diff
@@ -0,0 +1,26 @@
+"""
+This utility would invoke pyyaml to load and dump the DRY yaml file
+with both merge key anchors and shared dictionary elements. 
+Pyyaml would handle the merge key elements properly with overrides.
+However, Pyyaml would not expand anchored shared dictionary elements.
+We then use the ruamel commandline tool "merge-expan" to make sure shared
```
zacharis278: nit. typo on "merge-expand"

zacharis278 (approved):
💯

Makefile
```diff
@@ -112,7 +112,7 @@ detect_changed_source_translations: ## check if translation files are up-to-date
 validate_translations: fake_translations detect_changed_source_translations ## install fake translations and check if translation files are up-to-date
 
 api_generated: ## generates an expanded verison of api.yaml for consuming tools that cannot read yaml anchors
-	yaml merge-expand api.yaml .api-generated.yaml
+	python yml_merge.py api.yaml .api-generated.yaml
 
 validate_api_committed: ## check to make sure any api.yaml changes have been committed to the expanded document
 	bash -c "diff .api-generated.yaml <(yaml merge-expand api.yaml -)"
```
zacharis278: `validate_api_committed` needs to use your new task as well

schenedx: Yes. Found that out. Could you check if what I am doing is still good?

kdmccormick (commented):
Let's keep the root directory clean. Please move `yml_merge.py` to `scripts/`.

Also, to make sure that the scripts in `scripts/` get their code style checked, can you change in the Makefile:

```Makefile
quality:
    pycodestyle registrar *.py
    ...
```
to:
```Makefile
quality:
    pycodestyle registrar scripts
    ...
```

Other than that, looks good!

yml_merge.py
```diff
@@ -0,0 +1,34 @@
+"""
```
iloveagent57: We have a `scripts/` directory, can this live in there?

kdmccormick:
One more nit: `yaml_merge.py` would probably be a more appropriate name

yml_merge.py
```diff
@@ -0,0 +1,34 @@
...
+        output_path = '/tmp/generated.yml'
+
+    with open(output_path, 'w') as wp:
+        yaml.dump(data, wp)
+
+    # Now, use ruamel yaml to merge-expand
+    os.system('yaml merge-expand {0} {1}'.format(output_path, sys.argv[2]))
+
+    if output_path != sys.argv[2]:
+        os.remove(output_path)
```
iloveagent57: Is it safe to remove this file if the `os.system(...)` call fails?

scripts/yml_merge.py
```diff
@@ -0,0 +1,34 @@
+"""
+This utility would invoke pyyaml to load and dump the DRY yaml file
```
iloveagent57: what's a "DRY" yaml file?

yml_merge.py
schenedx: Done

yml_merge.py
schenedx: Yes in the grand scheme of things. The worst would be deleting the `.api-generated.yaml`, which we have in github so we don't necessarily lose it. 
All other problems, we can easily notice and fix locally.

scripts/yml_merge.py
schenedx: I made it up. It's the file with anchors and alias we need this process for. What's your name for it?

kdmccormick: "Don't Repeat Yourself"?

kdmccormick (approved):
Send it :+1:

PR #73, opened 2019-04-23
updates for integration with api-gateway

zacharis278:
This is the api file our deployment is currently pointing at.  Once this is merged I'll update api-manager to point to the new commit hash on master

1. api-gateway integration fixes / additions
2. removed 'example' attributes from models.  While they are valid swagger 2.0, api-gateway does not support them.

### DIFF
--- a/.api-generated.yaml
+++ b/.api-generated.yaml
@@ -6,7 +6,9 @@ course_enrollment_body:
   name: body
   required: true
   schema:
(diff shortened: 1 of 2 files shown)


jansenk:
> 2. removed 'example' attributes from models.  While they are valid swagger 2.0, api-gateway does not support them.

this is so frustrating >< amazon you have no excuse for this

PR #75, opened 2019-04-25
Add segment tracking for all V1-mock views

schenedx:
## Description

This PR is to make sure all the request to v1-mock views will have segment tracking.

@edx/masters-neem Please review

### DIFF
--- a/registrar/apps/api/segment.py
+++ b/registrar/apps/api/segment.py
@@ -5,6 +5,8 @@ import logging
 import analytics
 from django.conf import settings
 
+from registrar.apps.core.utils import get_user_organizations
+
 logger = logging.getLogger(__name__)
 
 
@@ -21,3 +23,14 @@ def track(
         analytics.track(user_id, event, properties, context, timestamp, anonymous_id, integrations)
     else:
         logger.debug("{{{}, {}}} not tracked because SEGMENT_KEY not set".format(user_id, event))
+
+
+def get_tracking_properties(user, **kwargs):
+    """
+    Helper function to construct the properties for tracking events
+    """
+    user_orgs = get_user_organizations(user)
+    property_dict = kwargs.copy()
+    property_dict['user_organizations'] = [org.name for org in user_orgs]
+
+    return property_dict
--- a/registrar/apps/api/v1_mock/tests/test_views.py
+++ b/registrar/apps/api/v1_mock/tests/test_views.py
@@ -13,34 +13,74 @@ from registrar.apps.api.v1_mock.data import (
     invoke_fake_program_enrollment_listing_job,
 )
 from registrar.apps.core.permissions import OrganizationReadMetadataRole
-from registrar.apps.core.tests.factories import GroupFactory, UserFactory
+from registrar.apps.core.tests.factories import (
+    GroupFactory,
+    UserFactory,
+    OrganizationFactory,
+    OrganizationGroupFactory
+)
 
 
 class MockAPITestMixin(AuthRequestMixin):
     """ Base mixin for tests for the v1_mock API. """
     api_root = '/api/v1-mock/'
+    event = None
 
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        cls.user = UserFactory()
+        cls.user_org = OrganizationFactory()
+        cls.user_org_group = OrganizationGroupFactory(organization=cls.user_org)
+        cls.user = UserFactory(groups=[cls.user_org_group])
         cls.admin_read_metadata_group = GroupFactory(
             name='admin_read_metadata',
             permissions=OrganizationReadMetadataRole.permissions,
         )
         cls.admin_user = UserFactory(groups=[cls.admin_read_metadata_group])
+        cls.patcher = mock.patch('registrar.apps.api.segment.track', autospec=True)
+
+    def setUp(self):
+        super().setUp()
+        self.mock_track = self.patcher.start()
+
+    @property
+    def path(self):
+        return self.api_root + self.path_suffix
+
+    def assert_segment_tracking(self, user=None, **kwargs):
+        """
+        Make sure the segment events are created when the request is made
+        """
+        properties = kwargs.copy()
+
+        if 'user_organizations' not in properties:
+            properties['user_organizations'] = [self.user_org.name]
+
+        if not user:
+            user = self.user
+
+        self.mock_track.assert_called_once_with(
+            user.username,
+            self.event,
+            properties
+        )
+
+    def tearDown(self):
+        super().tearDown()
+        self.patcher.stop()
 
 
-class MockJobTestMixin(object):
+class MockJobTestMixin(MockAPITestMixin):
     """ Mixin for testing the results of a job """
 
-    def assert_job_result(self, job_url, expected_state, expected_path):
+    def assert_job_result(self, job_id, job_url, expected_state, expected_path):
         """
         Gets the job at ``job_url``. Asserts:
          * Response is 200
          * Job dict has 'created' key
          * Provided ``expected_state`` matches job's
          * Provided ``expected_path`` is substring of job's result
+         * Segment event is created for job retrival
         """
         job_response = self.get(job_url, self.user)
         self.assertEqual(200, job_response.status_code)
@@ -51,6 +91,11 @@ class MockJobTestMixin(object):
             self.assertIn(expected_path, job_data['result'])
         else:
             self.assertIsNone(job_data['result'])
+        self.event = 'registrar.v1_mock.get_job_status'
+        self.assert_segment_tracking(
+            job_id=job_id,
+            job_state=job_data['state'],
+        )
 
 
 @ddt.ddt
@@ -59,19 +104,30 @@ class MockProgramListViewTests(MockAPITestMixin, APITestCase):
 
     method = 'GET'
     path = 'programs'
+    event = 'registrar.v1_mock.list_programs'
 
     def test_list_all_unauthorized(self):
         response = self.get(self.path, self.user)
         self.assertEqual(response.status_code, 403)
+        self.assert_segment_tracking(permission_failure='no_org_key')
 
     def test_list_all_admin_user(self):
         response = self.get(self.path, self.admin_user)
         self.assertEqual(response.status_code, 200)
         self.assertEqual(len(response.data), len(FAKE_PROGRAMS))
+        self.assert_segment_tracking(
+            user=self.admin_user,
+            user_organizations=[],
+        )
 
     def test_list_org_unauthorized(self):
-        response = self.get(self.path + '?org=u-perezburgh', self.user)
+        org_key = 'u-perezburgh'
+        response = self.get(self.path + '?org=' + org_key, self.user)
         self.assertEqual(response.status_code, 403)
+        self.assert_segment_tracking(
+            organization_filter=org_key,
+            permission_failure='no_org_permission',
+        )
 
     def test_list_org_no_perm_but_admin(self):
         """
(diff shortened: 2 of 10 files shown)


registrar/apps/api/segment.py
```diff
@@ -21,3 +23,34 @@ def track(
...
+            pass
+    return user_organizations
+
+
+def get_tracking_properties(user, **kwargs):
+    """
+    Helper function to construct the properties for tracking events
+    """
+    user_orgs = get_user_organizations(user)
+    property_dict = {
```
kdmccormick: This can be done in fewer lines:
```python
property_dict = kwargs.copy()
property_dict['user_organizations'] = ...
```

registrar/apps/api/segment.py
```diff
@@ -21,3 +23,34 @@ def track(
         analytics.track(user_id, event, properties, context, timestamp, anonymous_id, integrations)
     else:
         logger.debug("{{{}, {}}} not tracked because SEGMENT_KEY not set".format(user_id, event))
+
+
+def get_user_organizations(user):
```
kdmccormick: This is a great utility function! I can see it being used other places in the codebase in the future. Maybe `registrar/apps/core/utils.py` would be a good place for it?

registrar/apps/api/v1_mock/tests/test_views.py
```diff
@@ -13,42 +13,77 @@
...
 
     def test_unauthenticated(self):
         response = self.get(self.path, None)
         self.assertEqual(response.status_code, 401)
 
+    def assert_segment_tracking(self, user=None, **kwargs):
+        """
+        Make sure the segment events are created when the request is made
+        """
+        properties = {}
```
kdmccormick: One liner: `properties = kwargs.copy()`

registrar/apps/api/v1_mock/tests/test_views.py
```diff
@@ -13,42 +13,77 @@
...
+        for key, value in kwargs.items():
+            properties[key] = value
+
+        if properties.get('user_organizations', None) is None:
```
kdmccormick: More Pythonic: `if 'user_organizations' not in properties:`

registrar/apps/api/v1_mock/tests/test_views.py
```diff
@@ -59,26 +94,44 @@ def assert_job_result(self, job_url, expected_state, expected_path):
...
+        )
 
     def test_list_org_unauthorized(self):
-        response = self.get(self.path + '?org=u-perezburgh', self.user)
+        org_key = 'u-perezburgh'
+        response = self.get(self.path + '?org=' + org_key, self.user)
         self.assertEqual(response.status_code, 403)
+        self.assert_segment_tracking(
+            organization_filter=org_key,
+            permission_failure='no_permission_on_org'
```
kdmccormick: General style nit: trailing commas on these multi-line function calls are recommended in the edX style guide

registrar/apps/api/v1_mock/tests/test_views.py
```diff
@@ -103,13 +156,17 @@ def test_success(self, org_key, num_programs):
         response = self.get(self.path + '?org={}'.format(org_key), self.user)
         self.assertEqual(response.status_code, 200)
         self.assertEqual(len(response.data), num_programs)
+        self.assert_segment_tracking(
```
kdmccormick: General style nit: Lots of these calls can be on one line

registrar/apps/api/v1_mock/views.py
```diff
@@ -20,6 +20,7 @@
     HTTP_422_UNPROCESSABLE_ENTITY,
 )
 
+import registrar.apps.api.segment as segment
```
kdmccormick: Nit: `from registrar.apps.api import segment` is more consistent with the rest of our code.

registrar/apps/api/v1_mock/views.py
```diff
@@ -353,20 +430,58 @@ def get_serializer_class(self):
...
+            segment.get_tracking_properties(
+                self.request.user,
+                program_key=self.program.key,
+                course_key=self.course.key if self.course else None,
+            )
+        )
         if not self.program.managing_organization.enrollments_readable:
             raise PermissionDenied()
 
+        if self.course is None:
```
kdmccormick: I see you moved `if self.course is None: raise Http404()` out of `def course()`, presumably so that we can raise the `segment.track` event even if the course key is wrong.

Is there a reason we'd want to do this for invalid course keys, but not for invalid program keys?

registrar/apps/api/v1_mock/views.py
```diff
@@ -91,20 +92,52 @@ def check_permissions(self, request):
         if not self.org_key and global_read_metadata_perm(request.user):
             return
         elif not self.org_key:
+            segment.track(
```
kdmccormick: The `segment.track` calls, while important, take up a lot of space in this module and create a lot of noise. It would be nice if we could factor lots of the common code out into a utility function in a mixin.

For example, every `segment.track` call includes `self.request.user.username`, the name of the event (which always starts with `'registrar.v1_mock.'`, and a call to `get_tracking_properties` with `self.request.user` as the first argument. I could imagine being able to just call `self.segment_track('list_program', permission_failure='no_org_key')`

registrar/apps/api/v1_mock/tests/test_views.py
```diff
@@ -13,42 +13,77 @@
...
         )
         self.admin_user = UserFactory(groups=[self.admin_read_metadata_group])
+        self.patcher = mock.patch('registrar.apps.api.segment.track', autospec=True)
+        self.mock_track = self.patcher.start()
 
     def test_unauthenticated(self):
         response = self.get(self.path, None)
         self.assertEqual(response.status_code, 401)
 
+    def assert_segment_tracking(self, user=None, **kwargs):
```
kdmccormick, line 50: Nice general test structure :+1:

registrar/apps/api/v1_mock/views.py
```diff
@@ -295,15 +342,45 @@ def get_serializer_class(self):
...
+        segment.track(
+            self.request.user.username,
+            'registrar.v1_mock.get_program_enrollment',
+            segment.get_tracking_properties(
+                self.request.user,
+                program_key=self.program.key
+            )
+        )
         if not self.program.managing_organization.enrollments_readable:
             raise PermissionDenied()
```
kdmccormick, line 423: We fire segment events for some 4xx errors, but not for all of them. Should we standardize on firing events for all 403s/422s/etc.?

kdmccormick (changes requested):
Looking good in general. I have a few stylistic suggestions and a couple functionality questions.

registrar/apps/api/v1_mock/tests/test_views.py
schenedx: Ah, I was doing `if not hasattr(properties, 'user_organizations'):`. Then I realized I was barking at the wrong tree.
Thank you for this suggestion. Make sense.

registrar/apps/api/v1_mock/views.py
schenedx: Yep. We should fire events for these as well. Just a miss on my part.

registrar/apps/api/v1_mock/views.py
schenedx: We should do this for invalid program as well. Let me check to see how to do that.

kdmccormick: One suggestion that would allow you to keep the `raise Http404()`s within `def course()` and `def program()`:

Define on every class a mapping from methods to event names, for example: 
`events = {'PATCH': 'patch_program_enrollments', 'POST': 'post_program_enrollments, 'GET': 'get_program_enrollments'}`.

Then, within `def program()` you could do `segment.track(user, self.events[self.request.method], ...)` right before a 404 is raised.

schenedx:
@kdmccormick I took in all your suggestions and rebased. Please review again.

registrar/apps/api/v1_mock/tests/test_views.py
```diff
@@ -13,34 +13,74 @@
...
         cls.admin_read_metadata_group = GroupFactory(
             name='admin_read_metadata',
             permissions=OrganizationReadMetadataRole.permissions,
         )
         cls.admin_user = UserFactory(groups=[cls.admin_read_metadata_group])
+        cls.patcher = mock.patch('registrar.apps.api.segment.track', autospec=True)
+
+    def setUp(self):
+        super().setUp()
+        self.mock_track = self.patcher.start()
```
kdmccormick, line 44: This can be rolled into `setUpClass`, and `tearDown` could be `tearDownClass`

registrar/apps/api/v1_mock/views.py
```diff
@@ -52,7 +56,41 @@ def global_read_metadata_perm(user):
     return user.has_perm(perms.ORGANIZATION_READ_METADATA)
 
 
-class MockProgramListView(ListAPIView):
+class SegmentTrackViewMixin(object):
```
kdmccormick, line 59: Nice! When we do the V1 events, let's try to remember to pull this mixin up a package level and use it in both API versions.

registrar/apps/api/v1_mock/views.py
```diff
@@ -409,6 +510,15 @@ class MockJobStatusRetrieveView(RetrieveAPIView):
     def get_object(self):
         job_id = self.kwargs.get('job_id')
         job_status = get_fake_job_status(job_id, self.request.build_absolute_uri)
+        segment.track(
```
kdmccormick: Can this call be done with `self.segment_track`?

registrar/apps/api/v1_mock/tests/test_views.py
```diff
@@ -13,34 +13,74 @@
...
+        properties = kwargs.copy()
 
+        if 'user_organizations' not in properties:
+            properties['user_organizations'] = [self.user_org.name]
 
-class MockJobTestMixin(object):
+        if not user:
+            user = self.user
+
+        self.mock_track.assert_called_with(
```
kdmccormick: Would it be fair to test `assert_called_once_with` instead? I feel like we should only be emitting one of these events per request.

Doing so causes about 30 test failures, all on tests in classes `Mock...EnrollmentTests`. I think that's because the way eventing on `validate_and_echo_statuses` is currently implemented causes multiple events in certain cases.

registrar/apps/api/v1_mock/views.py
```diff
@@ -295,18 +360,28 @@ def get_serializer_class(self):
             return ProgramEnrollmentModificationRequestSerializer
 
     def post(self, request, *args, **kwargs):
+        """
+        The view to handle the POST method on program enrollments
+        """
+        self.segment_track(program_key=self.program.key)
```
kdmccormick: This will cause multiple events to be fired if there is a failure (this event and the failure event). I recommend emitting the event from within `validate_and_echo_statuses`.

kdmccormick (changes requested):
Great progress; I think this is almost there. However, there are places where multiple events are being emitted.

registrar/apps/api/v1_mock/tests/test_views.py
schenedx: This cannot. I tried. We need to patch each time a test is run. Otherwise I think the call counter in the mock won't reset.

registrar/apps/api/v1_mock/views.py
schenedx: Yes. Exactly what I was thinking too.

schenedx:
@kdmccormick Took all of your feedback in. Please review again.

registrar/apps/api/v1_mock/tests/test_views.py
kdmccormick: Ah, makes sense. Sounds good.

kdmccormick (approved):
:+1:

I have one suggestion, but my approval stands whether or not you make this change: It might be good to include `program_key` and `course_key` (if available) in the 403 and 404 events.

schenedx:
> 👍
> 
> I have one suggestion, but my approval stands whether or not you make this change: It might be good to include `program_key` and `course_key` (if available) in the 403 and 404 events.

Took in your suggestion above. I am going to merge.

PR #76, opened 2019-04-29
PATCH program enrollments

zacharis278:

### DIFF
--- a/registrar/apps/api/v1/tests/test_views.py
+++ b/registrar/apps/api/v1/tests/test_views.py
@@ -87,7 +87,7 @@ class RegistrarAPITestCase(APITestCase):
 
     def mock_api_response(self, url, response_data, method='GET', response_code=200):
         responses.add(
-            getattr(responses, method),
+            getattr(responses, method.upper()),
             url,
             body=json.dumps(response_data),
             content_type='application/json',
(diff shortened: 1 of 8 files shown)


registrar/apps/api/v1/tests/test_views.py
```diff
@@ -338,8 +338,8 @@ def test_program_not_found(self):
         self.assertEqual(response.status_code, 404)
 
 
-class ProgramEnrollmentPostTests(RegistrarAPITestCase, AuthRequestMixin):
-    """ Tests for the POST /api/v1/programs/{program_key}/enrollments endpoint """
+class ProgramEnrollmentWriteTests(RegistrarAPITestCase, AuthRequestMixin):
```
schenedx, line 342: Instead of having both POST and PATCH tested in the same class, I rather seeing two different classes with a common mixin. This way, you don't need to run things twice in a test method all the time.
And you can put a lot of common data in the mixin.

registrar/apps/api/v1/views.py
```diff
@@ -349,18 +347,25 @@ def get(self, request, *args, **kwargs):
...
+        self.validate_enrollment_data(request.data)
+        program_uuid = self.program.discovery_uuid
+        response = write_program_enrollments(program_uuid, request.data)
+        return Response(response.json(), status=response.status_code)
 
+    def patch(self, request, program_key):
+        """ PATCH handler """
+        self.validate_enrollment_data(request.data)
         program_uuid = self.program.discovery_uuid
-        response = post_lms_program_enrollment(program_uuid, request.data)
```
schenedx, line 363: The 3 lines before this and one line after this are exactly the same. Maybe create a common handler to make it DRY?

schenedx (changes requested):
I think we can make the code change a bit more structured to read better?

registrar/apps/enrollments/data.py
```diff
@@ -35,17 +35,22 @@ def get_discovery_program(program_uuid, client=None):
...
     """
-    Enroll students in a program
+    Create or update program enrollments in the LMS.
 
     Returns:
         A HTTP response object that includes both response data and status_code
     """
     url = urljoin(settings.LMS_BASE_URL, 'api/program_enrollments/v1/programs/{}/enrollments/').format(program_uuid)
 
+    if update is True:
```
kdmccormick:
```python
if update:
    ...
```

Alternatively, `method = 'PATCH' if update else 'POST'`

registrar/apps/api/v1/tests/test_views.py
zacharis278: I updated this so the tests are parameterized since POST and PATCH tests are exactly the same.  I went with that over a mixin since otherwise all of the actual test definitions would be in the mixin instead.

schenedx (approved):
👍 LGTM

PR #77, opened 2019-04-29
Use new API_ROOT setting to build absolute URLs

kdmccormick:
@edx/masters-neem 
[EDUCATOR-4274](https://openedx.atlassian.net/browse/EDUCATOR-4274)

Configuration PR: https://github.com/edx/configuration/pull/5102
Internal configuration PR: https://github.com/edx/edx-internal/pull/849

### DIFF
--- a/registrar/apps/api/utils.py
+++ b/registrar/apps/api/utils.py
@@ -1,0 +1,50 @@
+""" Miscellaneous utilities for API. """
+
+from django.conf import settings
+from django.urls import reverse
+
+
+def build_absolute_api_url(url_name, **kwargs):
+    """
+    Build a path to a Registrar API endpoint, and then make it into an
+    absolute, externally-accessible API URL.
+
+    Convenience wrapper around to_absolute_api_url.
+
+    Arguments:
+        url_name (str): name of URL as configured in urls.py.
+        kwargs (dict): URL kwargs for call to `reverse`.
+
+    Returns: str
+    """
+    return to_absolute_api_url(reverse(url_name, kwargs=kwargs))
+
+
+def to_absolute_api_url(path, *more_paths):
+    """
+    Make a path into an absolute, externally-accessible API URL.
+
+    Attempts to gracefully handle missing/duplicate forward slashes,
+    while preserving whether or not there is a trailing slash.
+
+    Arguments:
+        path (str): Path root. Must begin with /api/
+        more_paths (list[str]): Paths to be appended to `path`, in order.
+
+    Example:
+        to_absolute_api_url('api/v1', 'programs', 'abcd/enrollments/')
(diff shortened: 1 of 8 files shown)


registrar/apps/api/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    Convert a path to a resource on Registrar to an absolute,
+    externally-accessible URL.
+
+    Arguments:
+        path (str)
+
+    Returns: str
+    """
+    return settings.URL_ROOT.strip('/') + '/' + path.strip('/')
```
iloveagent57: Nit: you could make this a little more precise with `rstrip` and `lstrip`:
```python
return settings.URL_ROOT.rstrip('/') + '/' + path.lstrip('/')
```
That way you don't have to worry about stripping off a trailing "/" (I don't know if that's even a consideration, here, though).

registrar/apps/api/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    absolute, externally-accessible URL.
+
+    Arguments:
+        url_name (str): name of URL as configured in urls.py.
+        kwargs (dict): URL kwargs for call to `reverse`.
+
+    Returns: str
+    """
+    raw_url = reverse(url_name, kwargs=kwargs)
+    if raw_url.startswith('/api/'):
```
iloveagent57: NIT: you know that `raw_url` starts with "/api/" inside this block, so why not do:
```python
return settings.API_ROOT.rstrip('/') + '/' + raw_url.lstrip('/api/')
```

iloveagent57 (commented):
I know this is a draft, just a couple minor suggestions about string munging.

registrar/apps/api/utils.py
kdmccormick: Ended up rewriting this part, but I made sure to preserve the presence of a trailing slash this time.

registrar/apps/api/utils.py
```diff
@@ -0,0 +1,50 @@
...
+            => 'https://{API_ROOT}/api/v1/programs/123'
+
+    Returns: str
+    """
+    PREFIX = '/api/'
+    if not path.startswith(PREFIX):
+        raise ValueError(
+            'Cannot make API URL for raw URL that does not begin with ' + PREFIX
+        )
+    path_parts = [settings.API_ROOT, path[len(PREFIX):]] + list(more_paths)
```
schenedx, line 47: Please write a unit test for this block of logic. It's difficult to access the different case you are trying to handle by reading this code alone.

schenedx (commented):
I think we just need a set of unit test, and this is good to go.

zacharis278:
lgtm pending some tests.  However, if we're using /media to serve up the results there's nothing exposing that path in api.yaml.

zacharis278 (approved):

PR #79, opened 2019-05-02
Include curriculum_uuid with program enrollments

zacharis278:
### [EDUCATOR-4283](https://openedx.atlassian.net/browse/EDUCATOR-4283)

1. Send curriculum_uuid to LMS when creating or updating program enrollments
2. Added caching of discovery programs
3. Fix POST/PATCH to use base class instead of ddt

Note: It turns out discovery does not expose the created date on curricula, for now this will use the first active curriculum in a program.

### DIFF
--- a/registrar/apps/api/v1/tests/test_views.py
+++ b/registrar/apps/api/v1/tests/test_views.py
@@ -211,25 +211,29 @@ class ProgramCourseListViewTests(RegistrarAPITestCase, AuthRequestMixin):
         user = self.edx_admin if is_staff else self.hum_admin
 
         program_data = {
-            'curricula': [{
-                'courses': [{
-                    'course_runs': [
-                        {
-                            'key': '0001',
-                            'uuid': '123456',
-                            'title': 'Test Course 1',
-                            'marketing_url': 'https://humanities-college.edx.org/masters-in-english/test-course-1',
-                        }
-                    ],
-                }],
-            }],
+            'curricula': [
+                {
+                    'is_active': False,
+                    'courses': []
+                },
+                {
+                    'is_active': True,
+                    'courses': [{
+                        'course_runs': [
+                            {
+                                'key': '0001',
+                                'uuid': '123456',
+                                'title': 'Test Course 1',
+                                'marketing_url': 'https://humanities-college.edx.org/masters-in-english/test-course-1',
+                            }
+                        ],
+                    }]
+                },
+            ]
         }
 
-        program_uuid = self.english_program.discovery_uuid
-        discovery_url = urljoin(settings.DISCOVERY_BASE_URL, 'api/v1/programs/{}/').format(program_uuid)
-        self.mock_api_response(discovery_url, program_data)
-
-        response = self.get('programs/masters-in-english/courses', user)
+        with mock.patch('registrar.apps.api.v1.views.get_discovery_program', return_value=program_data):
+            response = self.get('programs/masters-in-english/courses', user)
 
         self.assertEqual(response.status_code, 200)
         self.assertListEqual(
@@ -252,17 +256,35 @@ class ProgramCourseListViewTests(RegistrarAPITestCase, AuthRequestMixin):
 
         program_data = {
             'curricula': [{
+                'is_active': True,
                 'courses': [{
                     'course_runs': []
                 }]
             }]
         }
 
-        program_uuid = self.english_program.discovery_uuid
-        discovery_url = urljoin(settings.DISCOVERY_BASE_URL, 'api/v1/programs/{}/').format(program_uuid)
-        self.mock_api_response(discovery_url, program_data)
+        with mock.patch('registrar.apps.api.v1.views.get_discovery_program', return_value=program_data):
+            response = self.get('programs/masters-in-english/courses', user)
 
-        response = self.get('programs/masters-in-english/courses', user)
+        self.assertEqual(response.status_code, 200)
+        self.assertListEqual(response.data, [])
+
+    @mock_oauth_login
+    @responses.activate
+    def test_get_program_with_no_active_curriculum(self):
+        user = self.hum_admin
+
+        program_data = {
+            'curricula': [{
+                'is_active': False,
+                'courses': [{
+                    'course_runs': []
+                }]
+            }]
+        }
+
+        with mock.patch('registrar.apps.api.v1.views.get_discovery_program', return_value=program_data):
+            response = self.get('programs/masters-in-english/courses', user)
 
         self.assertEqual(response.status_code, 200)
         self.assertListEqual(response.data, [])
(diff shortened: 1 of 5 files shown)


registrar/apps/enrollments/data.py
```diff
@@ -31,8 +35,14 @@ def get_discovery_program(program_uuid, client=None):
...
     Fetches program data from discovery service, given a program UUID.
     """
-    url = urljoin(settings.DISCOVERY_BASE_URL, 'api/v1/programs/{}/').format(program_uuid)
-    return _make_request('GET', url, client).json()
+    program = cache.get(PROGRAM_CACHE_KEY_TPL.format(uuid=program_uuid))
+
+    if not program:
+        url = urljoin(settings.DISCOVERY_BASE_URL, 'api/v1/programs/{}/').format(program_uuid)
+        program = _make_request('GET', url, client).json()
+        cache.set(PROGRAM_CACHE_KEY_TPL.format(uuid=program_uuid), program)
```
kdmccormick: I know this is still a draft, but heads up: the default cache lifetime is 5 minutes. If you want it to last indefinitely, pass in `None` here, or set the default timeout to `None` in config.

zacharis278: TIL.  thanks

zacharis278:
@schenedx figured out what I was doing wrong with having a base mixin for POST/PATCH with `test_*` functions so I fixed that to not use DDT anymore

registrar/apps/api/v1/views.py
```diff
@@ -262,11 +262,15 @@ def get_queryset(self):
...
 
         # this make two temporary assumptions (zwh 03/19)
-        #  1. one curriculum per program
+        #  1. one *active* curriculum per program
         #  2. no programs are nested within a curriculum
         course_runs = []
         if curricula:
-            for course in curricula[0].get('courses') or []:
+            try:
+                curriculum = next(c for c in curricula if c['is_active'] is True)
```
schenedx: #NIT This is being invoked twice. I can see this being a utility function that can be used in the future?

registrar/apps/enrollments/data.py
```diff
@@ -2,15 +2,19 @@
...
 from django.conf import settings
+from requests.exceptions import HTTPError
+
 from edx_rest_api_client import client as rest_client
 
 from registrar.apps.enrollments.serializers import (
     ProgramEnrollmentSerializer,
 )
 
+PROGRAM_CACHE_KEY_TPL = 'program-{uuid}'
```
schenedx: Should this be moved to the constants.py file @jansenk merged in?

registrar/apps/enrollments/tests/test_data.py
```diff
@@ -91,3 +93,84 @@ def test_get_program_enrollments_500(self):
         responses.add(responses.GET, self.url, status=500)
         with self.assertRaises(HTTPError):
             get_program_enrollments(self.uuid)
+
+
+class GetDiscoveryProgramTestCase(TestCase):
+    """ Test get_discovery_program function """
+
+    program_uuid = uuid.uuid4().hex[0:10]
```
schenedx: Why do we only need the `hex` representation of the first 11 characters? Why not just take `uuid.uuid()` at its face value?

schenedx (approved):
👍 All my inline comments are non blocking questions or suggestions

registrar/apps/api/v1/views.py
zacharis278: opted to leave these separate since it's a temp workaround and the cases are _slightly_ different.

registrar/apps/enrollments/tests/test_data.py
zacharis278: overly liberal with copypasta, fixed

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -361,97 +370,122 @@ def student_enrollment(self, status, student_key=None):
...
+    def test_write_enrollment_payload_limit(self):
         req_data = [self.student_enrollment('enrolled')] * 26
 
-        response = getattr(self, method)('programs/masters-in-cs/enrollments/', req_data, self.stem_admin)
+        response = self.make_request('programs/masters-in-cs/enrollments/', req_data, self.stem_admin)
         self.assertEqual(response.status_code, 413)
 
 
+class ProgramEnrollmentPostTests(ProgramEnrollmentWriteMixin, RegistrarAPITestCase, AuthRequestMixin):
+    method = 'POST'
```
kdmccormick, line 498: Nice :+1: :+1:

registrar/apps/api/v1/views.py
kdmccormick: Nit: You can drop the `is True`; `c['is_active']` is already a bool. Same for the other appearance of this.

registrar/apps/api/v1/views.py
```diff
@@ -262,11 +262,15 @@ def get_queryset(self):
...
+            except StopIteration:
+                pass
```
kdmccormick, line 274: If we get a `StopIteration`, wouldn't `curriculum` be `None`? And then wouldn't the for loop raise an error?

registrar/apps/enrollments/data.py
iloveagent57: A minor suggestion about cache keys: a pattern I really like, especially for keeping lines shorter, is to use both a template like you have plus a helper function:
```python
def program_cache_key(program_uuid):
    return PROGRAM_CACHE_KEY_TPL.format(uuid=program_uuid)
```

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -352,6 +350,17 @@ def setUpClass(cls):
...
         ).format(program_uuid)
 
+        cls.program_curriculum_data = {
+            'curricula': [
+                {'uuid': 'inactive-curriculum-0000', 'is_active': False},
+                {'uuid': 'active-curriculum-0000', 'is_active': True}
+            ]
+        }
+
+    def make_request(self, path, data, user):
```
kdmccormick: `AuthRequestMixin.request` is available to call directly.

registrar/apps/enrollments/tests/test_data.py
```diff
@@ -91,3 +93,84 @@ def test_get_program_enrollments_500(self):
...
+    @responses.activate
+    def test_get_discovery_program_error(self):
+        """
+        Verify if an error occurs when requesting discovery data an exception is
+        thown and the cache is not polluted.
+        """
+        responses.add(
+            responses.GET,
+            self.discovery_url,
+            json={'message': 'everything is broken'},
```
iloveagent57, line 162: https://en.wikipedia.org/wiki/Planet_Telex

iloveagent57 (approved):
This LGTM, just one very minor suggestion.

kdmccormick (changes requested):
Looking really good. Just one suggestion, one question, and one nit.

registrar/apps/api/v1/views.py
zacharis278: good call

kdmccormick (approved):

PR #80, opened 2019-05-03
Add the logging for v1_mock api endpoints

schenedx:
## Description

@edx/masters-neem  Please help review
This is the change to add logging to the API endpoints for debugging from splunk. It's largely riding on the segment tracking infrastructure.

### DIFF
--- a/registrar/apps/api/constants.py
+++ b/registrar/apps/api/constants.py
@@ -2,3 +2,4 @@
 Defines constants used by the registrar api.
 """
 ENROLLMENT_WRITE_MAX_SIZE = 25
+TRACKING_CATEGORY = 'Registrar API'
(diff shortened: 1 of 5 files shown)


iloveagent57 (approved):
This looks good to me, but you've got an assertion error in `test_org_not_found`...I can't spot the difference:
```
E   AssertionError: Expected call: info('%s invoked on Registrar with properties %s', 'registrar.v1_mock.list_programs', '{"org_key": "antarctica-tech", "failure": "org_not_found", "user_organizations": ["Test Origanization 14"], "category": "Registrar API"}')
E   Actual call: info('%s invoked on Registrar with properties %s', 'registrar.v1_mock.list_programs', '{"failure": "org_not_found", "org_key": "antarctica-tech", "user_organizations": ["Test Origanization 14"], "category": "Registrar API"}')
```
Once you fix that, ship it.

iloveagent57:
Oh, it's because you're comparing a json-serialized string version of a dictionary, so the order actually matters.  You should probably de-serialize into a dictionary and then compare.  Also, you have and extra 'i' in "Origanization".

PR #81, opened 2019-05-03
Implement V1 course enrollment GET

kdmccormick:
[EDUCATOR-4110](https://openedx.atlassian.net/browse/EDUCATOR-4110)
@edx/masters-neem 

This PR adds the asynchronous course-enrollment GET endpoint to the Registrar V1 API.

Currently, we cache Discovery program data as it comes: a JSON blob. Then, we extract necessary data and do validation on it per-request. In this PR, I propose extracting and validating that data *before* caching it. That logic is abstracted by the new `DiscoveryProgram` class in data.py. My reasoning is that we will eventually use the Discovery data in every V1 endpoint, so for the purposes of DRY, we should validate it all in one place. Additionally, by having a class with definite attributes, future developers will know what data to expect in the cache (instead of having to trawl through Discovery). Happy to hear any thoughts on this.

### DIFF
--- a/registrar/apps/api/v1/mixins.py
+++ b/registrar/apps/api/v1/mixins.py
@@ -1,0 +1,189 @@
+"""
+Mixins for the public REST API.
+"""
+from collections.abc import Iterable
+
+from django.core.exceptions import (
+    ImproperlyConfigured,
+    PermissionDenied,
+)
+from django.http import Http404
+from django.shortcuts import get_object_or_404
+from django.urls import resolve
+from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
+from guardian.shortcuts import get_objects_for_user
+from rest_framework.authentication import SessionAuthentication
+from rest_framework.exceptions import NotAuthenticated
+from rest_framework.response import Response
+from rest_framework.status import HTTP_202_ACCEPTED
+
+from registrar.apps.api.serializers import JobAcceptanceSerializer
+from registrar.apps.api.utils import build_absolute_api_url
+from registrar.apps.enrollments.models import Program
+from registrar.apps.core.jobs import start_job
+from registrar.apps.enrollments.data import DiscoveryProgram
+
+
+class AuthMixin(object):
+    """
+    Mixin providing AuthN/AuthZ functionality for all our views to use.
+
+    This mixin overrides `APIView.check_permissions` to use Django Guardian.
+    It replicates, to the extent that we require, the functionality of
+    Django Guardian's `PermissionRequiredMixin`, which unfortunately doesn't
+    play nicely with Django REST Framework.
+    """
+    authentication_classes = (JwtAuthentication, SessionAuthentication)
+    permission_required = []
+    raise_404_if_unauthorized = False
+
+    def get_permission_object(self):
+        """
+        Get an object against which required permissions will be checked.
+
+        If None, permissions will be checked globally.
+        """
+        return None
+
+    def get_permission_required(self, _request):
+        """
+        Gets permission(s) to be checked.
+
+        Must return a string or an iterable of strings.
+        Can be overridden in subclass.
+        Default to class-level `permission_required` attribute.
+        """
+        return self.permission_required
+
+    def check_permissions(self, request):
+        """
+        Check that the authenticated user can access this view.
+
+        Ensure that the user has all of the permissions specified in
+        `permission_required` granted on the object returned by
+        `get_permission_object`. If not, an HTTP 403 (or, HTTP 404 if
+        `raise_404_if_unauthorized` is True) is raised.
+
+        Overrides APIView.check_permissions.
+        """
+        if resolve(request.path_info).url_name == 'api-docs':
+            self.check_doc_permissions(request)
+            return
+
+        if not request.user.is_authenticated:
+            raise NotAuthenticated()
+
+        required = self.get_permission_required(request)
+        if isinstance(required, str):
+            required = [required]
+        elif isinstance(required, Iterable):
+            required = list(required)
+        else:
+            raise ImproperlyConfigured(
+                'get_permission_required must return string or iterable; ' +
+                'returned {}'.format(required)
+            )
+
+        if all(request.user.has_perm(perm) for perm in required):
+            return
+        obj = self.get_permission_object()
+        if obj and all(request.user.has_perm(perm, obj) for perm in required):
+            return
+        if self.raise_404_if_unauthorized:
+            raise Http404()
+        else:
+            raise PermissionDenied()
+
+    def check_doc_permissions(self, request):
+        """
+        Check whether the endpoint being requested should show up in the
+        Swagger UI.
+
+        When loading /api-docs/, Swagger does `check_permissions` on all
+        API endpoints in order to decide which ones to show to the user.
+        However, we assign permissions on a per-Oranization-instance
+        basis using Guardian, whereas /api-docs/ is Organization-agnostic.
+
+        To compensate for this, we handle permission checks coming from
+        /api-docs/ differently: we simply check if the user has the appropriate
+        permission on *any* Organization instance.
+        """
+        if not get_objects_for_user(request.user, self.permission_required):
+            raise PermissionDenied()
+
+
+class ProgramSpecificViewMixin(AuthMixin):
+    """
+    A mixin for views that operate on or within a specific program.
+
+    Provides a `program` property. On first access, the property is loaded
+    based on the `program_key` URL parameter, and cached for subsequent
+    calls. This avoids redundant database queries between `get_object/queryset`
+    and `get_permission_object`.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._program = None
+
+    @property
+    def program(self):
+        """
+        The program specified by the `program_key` URL parameter.
+        """
+        if not self._program:
+            program_key = self.kwargs['program_key']
+            self._program = get_object_or_404(Program, key=program_key)
+        return self._program
+
+    def get_permission_object(self):
+        """
+        Returns an organization object against which permissions should be checked.
+        """
+        return self.program.managing_organization
+
+
+class CourseSpecificViewMixin(ProgramSpecificViewMixin):
+    """
+    A mixin for views that operate on or within a specific program course run.
+
+    In addition to the functionality provided in `ProgramSpecificViewMixin`,
+    this mixin provides a `validate_course_id` function, which confirms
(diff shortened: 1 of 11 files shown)


registrar/apps/api/v1/mixins.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def validate_course_id(self):
+        """
+        Raises a 404 if the course run identified by the `course_id` path
+        parameter is not part of self.program.
+        """
+        course_id = self.kwargs['course_id']
+        program_uuid = self.program.discovery_uuid
+        discovery_program = DiscoveryProgram.get(program_uuid)
+        program_course_run_ids = [
```
iloveagent57: NIT: make this a set comprehension.

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -559,6 +566,92 @@ def test_program_not_found(self):
...
+        self.assertEqual(file_response.text, expected_contents)
+
+    @mock.patch.object(DiscoveryProgram, 'get', return_value=disco_program)
+    def test_permission_denied(self, _mock):
+        response = self.get(self.path, self.stem_admin)
+        self.assertEqual(response.status_code, 403)
+
+    @mock.patch.object(DiscoveryProgram, 'get', return_value=disco_program)
+    @ddt.data(
+        # Bad program
```
iloveagent57, line 639: Thank you for these comments, much easier to understand the test with them.

registrar/apps/api/v1/urls.py
```diff
@@ -37,6 +38,11 @@
         views.ProgramEnrollmentView.as_view(),
         name="program-enrollment",
     ),
+    url(
+        r'programs/{}/courses/{}/enrollments/$'.format(PROGRAM_KEY_PATTERN, COURSE_ID_PATTERN),
+        views.CourseEnrollmentView.as_view(),
+        name="program-enrollment",
```
iloveagent57: "program-course-enrollment", probably

registrar/apps/api/v1/views.py
```diff
@@ -338,18 +197,7 @@ def get(self, request, *args, **kwargs):
...
-        job_id = start_job(
-            self.request.user,
-            list_program_enrollments,
-            self.program.key,
-            file_format,
-        )
-        job_url = build_absolute_api_url('api:v1:job-status', job_id=job_id)
-        data = {'job_id': job_id, 'job_url': job_url}
-        return Response(JobAcceptanceSerializer(data).data, HTTP_202_ACCEPTED)
+        return self.invoke_job(list_program_enrollments, self.program.key)
```
iloveagent57, line 200: hooray one-line function!

registrar/apps/core/utils.py
```diff
@@ -30,3 +30,28 @@ def get_user_organizations(user):
...
+    Arguments:
+        items (list[dict])
+        field_names (tuple[str])
+        include_headers (bool)
+
+    Returns: str
+    """
+    def value_to_string(val):
+        """
+        Stringify a single value for CSV serialization.
```
iloveagent57: really only boolean values...

registrar/apps/core/utils.py
```diff
@@ -30,3 +30,28 @@ def get_user_organizations(user):
...
+        """
+        return str(val).lower() if isinstance(val, bool) else val
+
+    header = ",".join(field_names) + "\n" if include_headers else ""
+    body = "\n".join(
```
iloveagent57: You could use `csv.DictWriter` instead of this.  It's pretty tried and true.

registrar/apps/enrollments/data.py
```diff
@@ -11,37 +14,116 @@
...
+    Attributes:
+        * version (int)
+        * loaded (datetime): When data was loaded from Course Discovery
+        * uuid (str): Program UUID-4
+        * active_curriculum_uuid (str): UUID-4 of active curriculum.
+        * course_runs (list[DiscoveryCourseRun]):
+            Flattened list of all course runs in program
     """
-    program = cache.get(PROGRAM_CACHE_KEY_TPL.format(uuid=program_uuid))
+    class_version = 1  # Bump to invalidate cache.
```
iloveagent57: So is the idea here that, if we change something in the schema of this class, all of the old previously-cached objects should be ignored?

registrar/apps/enrollments/data.py
```diff
@@ -11,37 +14,116 @@
...
-        url = urljoin(settings.DISCOVERY_BASE_URL, 'api/v1/programs/{}/').format(program_uuid)
-        program = _make_request('GET', url, client).json()
-        cache.set(PROGRAM_CACHE_KEY_TPL.format(uuid=program_uuid), program, None)
+        Raises HTTPError if program is not cached and Discover returns error.
+        Raises ValidationError if program is not cached and Discovery returns
+            data in a format we don't like.
+        """
+        key = PROGRAM_CACHE_KEY_TPL.format(uuid=program_uuid)
+        program = cache.get(key)
+        if not (program and program.version == cls.class_version):
```
iloveagent57, line 69: If we have a program but it's version differs, should we delete the existing cache entry?

registrar/apps/enrollments/data.py
```diff
@@ -11,37 +14,116 @@
...
+        return cls.from_json(program_uuid, program_data)
+
+    @classmethod
+    def from_json(cls, program_uuid, program_data):
+        """
+        Builds a DiscoveryProgram instance from JSON data that has been
+        returned by the Course Discovery service.
+        """
+        try:
+            curriculum = next(
```
iloveagent57, line 99: I think this is the second or third time I've seen this chunk of code, can we make it a function?

registrar/apps/enrollments/data.py
```diff
@@ -11,37 +14,116 @@
...
+            )
+        active_curriculum_uuid = curriculum.get("uuid")
+        course_runs = [
+            DiscoveryCourseRun(
+                key=course_run.get('key'),
+                title=course_run.get('title'),
+                marketing_url=course_run.get('marketing_url'),
+            )
+            for course in curriculum.get("courses", [])
+            for course_run in course.get("course_runs", [])
```
iloveagent57, line 124: Hell yeah, with that nested comprehension!

registrar/apps/enrollments/tasks.py
```diff
@@ -54,7 +60,46 @@ def list_program_enrollments(self, job_id, user_id, program_key, file_format):
...
+        )
+        return
+
+    if file_format == 'json':
+        serialized = json.dumps(enrollments, indent=4)
+    elif file_format == 'csv':
+        serialized = serialize_course_run_enrollments_to_csv(enrollments)
+    else:
+        raise ValueError('Invalid file_format: {}'.format(file_format))
+    post_job_success(job_id, serialized, file_format)
```
iloveagent57, line 102: Nice, very clean and easy-to-read function body.

iloveagent57 (commented):
Nice job.  I think it's preferable to use the csv serialization that comes out of the box from the stdlib, unless you have a good reason not to.  Other than that and the question about cache invalidation, this looks good to go.

registrar/apps/api/v1/mixins.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+    A mixin for views that operate on or within a specific program course run.
+
+    In addition to the functionality provided in `ProgramSpecificViewMixin`,
+    this mixin provides a `validate_course_id` function, which confirms
+    that the course ID (provided in the `course_id` path parameter)
+    exists within a program. If it does not, a 404 is raised.
+    """
+
+    def validate_course_id(self):
```
iloveagent57, line 156: Also, if you wanted to eschew this, you could rely on the LMS API validating that the course exists.  That was my plan for POST/PATCH.

registrar/apps/api/v1/mixins.py
```diff
@@ -0,0 +1,189 @@
+"""
+Mixins for the public REST API.
```
zacharis278, line 2: definitely needed 👍

registrar/apps/enrollments/data.py
zacharis278, line 99: we probably want to call out in a comment here the assumption we're making regarding curricula composed of multiple programs since those will return nothing right now.

zacharis278 (approved):

registrar/apps/enrollments/data.py
kdmccormick: What you see is said function; this PR removes those other two chunks.

kdmccormick: Definitely. While refactoring I should've kept your comment.

registrar/apps/enrollments/tasks.py
kdmccormick: Thanks!

registrar/apps/enrollments/constants.py
```diff
@@ -1,6 +1,6 @@
 """ Constants for enrollments app """
 
-PROGRAM_CACHE_KEY_TPL = 'program-{uuid}'
+PROGRAM_CACHE_KEY_TPL = 'program:{uuid}'
```
schenedx, line 3: Why update the key template here?

registrar/apps/api/v1/mixins.py
kdmccormick: ~Per our conversation, I will rely on the LMS API validating that the course exists, and then change the LMS to do said validation~

schenedx (commented):
I don't have anymore comments than what you've already gotten. Good work

registrar/apps/api/v1/mixins.py
kdmccormick: We can't rely on the LMS, because this is an async endpoint. So that task would fail, but the user wouldn't see a 404.

registrar/apps/enrollments/constants.py
kdmccormick: Consistency with https://github.com/edx/registrar/blob/master/registrar/apps/api/v1_mock/data.py#L183

registrar/apps/enrollments/data.py
kdmccormick: Yep, good call.

kdmccormick: Actually, we do delete the existing entry, by nature of `cache.set(key, program, None)` overwriting it.

registrar/apps/enrollments/data.py
kdmccormick: Yep. I will make the comment more explicit and add a test that verifies that this works properly.

iloveagent57 (approved):

PR #83, opened 2019-05-07
Miscellaneous cleanup of code, models, etc.

kdmccormick:
@edx/masters-neem 

I figured deleting code and models would be safer to do before we have production users.

After merging this, we will deploy, then merge and deploy https://github.com/edx/registrar/pull/84


registrar/apps/enrollments/migrations/0006_remove_learner.py
```diff
@@ -0,0 +1,55 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.20 on 2019-05-06 22:19
+from __future__ import unicode_literals
```
schenedx: I would prefer you merged this file after this PR is merged and deployed first.
This way, we know this PR won't cause deployment problems.

schenedx (changes requested):
All looks good except we should keep the best practice on deleting columns or tables from our DB

registrar/apps/enrollments/migrations/0006_remove_learner.py
kdmccormick: @schenedx Good point. I put the migration file on a separate branch.

schenedx (approved):
👍

PR #85, opened 2019-05-07
Sort imports

kdmccormick:
@edx/masters-neem 

This PR:
1. Installs https://github.com/timothycrosley/isort
2. Adds `make isort` and `make validate_isort` targets
3. Adds `make validate_isort` to `make validate`, which adds it to Travis build
4. Runs `isort`

Steps 1-3 and step 4 are in two separate commits in order to make reviewing easier.

Rationale:
* I like import sorting because it makes it easier to tell what's what in a giant blob of imports. 
* We already mostly do it by hand.
* Theoretically, there should be fewer merge conflicts at the top of files, because there is a single deterministic way to order imports.
* edx-platform currently does it, so it must be a good idea.
* This is kinda wack: https://github.com/edx/registrar/blob/master/registrar/apps/core/tests/test_models.py#L9-L11

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,3 +2,13 @@
 ignore=E501,W504
 max-line-length = 120
 exclude=.git,settings,migrations,registrar/static,bower_components,registrar/wsgi.py
+
+[tool:isort]
+indent='    '
+line_length=80
+multi_line_output=3
+lines_after_imports=2
+include_trailing_comma=True
+skip=
+    settings
+    migrations
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,6 @@ install:
 script:
     - make validate_translations
     - make validate
-    - make pii_check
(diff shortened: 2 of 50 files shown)


setup.cfg
```diff
@@ -2,3 +2,13 @@
 ignore=E501,W504
 max-line-length = 120
 exclude=.git,settings,migrations,registrar/static,bower_components,registrar/wsgi.py
+
+[tool:isort]
+indent='    '
+line_length=80
```
zacharis278, line 8: if out pylint limit is 120 I'd think this would match

zacharis278 (approved):

setup.cfg
kdmccormick: I think we should aim to keep line lengths under 80 (but allow up to 120 for cases where short lines actually hurt readability)

PR #86, opened 2019-05-07
Make path regexes match only beginning of URL fragment

kdmccormick:
@edx/masters-neem 

@jansenk noticed that Registrar has some weird URL behavior.. http://localhost:18734/api/v1//api/v1/programs/ is a valid Registrar URL!

It's because some of our URL regex didn't begin with carets (^), meaning that the regex matches *anywhere* in the URL fragment, not just the beginning.

This URL fixes that, as well as a few tests that relied on that bug.... oops.


jansenk (approved):
lgtm

PR #87, opened 2019-05-07
add endpoint to clear program cache

jansenk:
Add endpoint to clear cache

### DIFF
--- a/registrar/apps/api/v1/mixins.py
+++ b/registrar/apps/api/v1/mixins.py
@@ -39,6 +39,7 @@ class AuthMixin(object):
     authentication_classes = (JwtAuthentication, SessionAuthentication)
     permission_required = []
     raise_404_if_unauthorized = False
+    staff_only = False
 
     def get_permission_object(self):
         """
@@ -76,6 +77,9 @@ class AuthMixin(object):
         if not request.user.is_authenticated:
             raise NotAuthenticated()
 
+        if self.staff_only and not request.user.is_staff:
+            self.unauthorized_response()
+
         required = self.get_permission_required(request)
         if isinstance(required, str):
             required = [required]
@@ -92,6 +96,9 @@ class AuthMixin(object):
         obj = self.get_permission_object()
         if obj and all(request.user.has_perm(perm, obj) for perm in required):
             return
+        self.unauthorized_response()
+
+    def unauthorized_response(self):
         if self.raise_404_if_unauthorized:
             raise Http404()
         else:
--- a/registrar/apps/api/v1/views.py
+++ b/registrar/apps/api/v1/views.py
@@ -43,7 +43,6 @@ from registrar.apps.enrollments.tasks import (
     list_program_enrollments,
 )
 
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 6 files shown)


kdmccormick:
Two thoughts on the URL --
* Maybe it should go under something like `api/internal/` instead of `api/v1/`, just to emphasize that it's not part of our public API
* From a semantic point of view, `DELETE ../cache` would make more sense than `DELETE ../flush-cache`

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -869,3 +871,90 @@ class ProgramCourseEnrollmentPostTests(ProgramCourseEnrollmentWriteMixin, Regist
 
 class ProgramCourseEnrollmentPatchTests(ProgramCourseEnrollmentWriteMixin, RegistrarAPITestCase, AuthRequestMixin):
     method = 'PATCH'
+
+
+@ddt.ddt
+class FlushCacheTests(RegistrarAPITestCase, AuthRequestMixin):
```
kdmccormick: Great tests! One more case I can think of -- flushing the same program twice still results in a 200.

registrar/apps/api/v1/views.py
```diff
@@ -332,3 +333,35 @@ def get_object(self):
...
+    def delete(self, request, program_key=None):
+        """
+        Clears a specific program from the cache, if specified.
+        Otherwise, clears the entire cache.
+        """
+        if program_key:
+            program_uuid = self.program.discovery_uuid
+            cache.delete(PROGRAM_CACHE_KEY_TPL.format(uuid=program_uuid))
+        else:
+            cache.clear()
```
kdmccormick: We only want to delete the program cache, not the entire cache. Right now we also use the cache for mock jobs, and in the future we might use it for more stuff. I recommend looping through `Program.objects.all()` to delete each cached program by UUID.

registrar/apps/api/v1/mixins.py
```diff
@@ -39,6 +39,7 @@ class AuthMixin(object):
     authentication_classes = (JwtAuthentication, SessionAuthentication)
     permission_required = []
     raise_404_if_unauthorized = False
+    staff_only = False
```
kdmccormick, line 42: :cool:

registrar/apps/api/v1/views.py
```diff
@@ -332,3 +333,35 @@ def get_object(self):
         except ObjectDoesNotExist:
             raise Http404()
         return job_status
+
+
+class FlushCacheView(ProgramSpecificViewMixin, APIView):
```
kdmccormick: I feel like the more explicit `FlushProgramCacheView` or `ProgramCacheView` would be preferable, in case we end up using Registrar's cache for stuff that doesn't get flushed by this view.

kdmccormick (changes requested):

registrar/apps/api/v1/views.py
jansenk: Oh, my bad. I looked through and was under the impression we didn't use the cache for jobs any more

kdmccormick (approved):
Good stuff. Approval pending a minor change -- a DELETE response with an empty body should return a 204 No Content.

jansenk:
> Good stuff. Approval pending a minor change -- a DELETE response with an empty body should return a 204 No Content.

What would the body contain?

PR #88, opened 2019-05-09
remove title and url from program [1/3]

jansenk:
Before I can change tests to not use title and url, I have to first change the title field to be nullable


kdmccormick (approved):
Approved, pending the migration file is named something other than `0007_auto_20190509_0250.py` :)

PR #89, opened 2019-05-09
v1 endpoints and annotations

zacharis278:
1. adds remaining v1 paths to api.yaml
2. add api-gateway attribute to all v1 endpoints


kdmccormick (approved):
:zap:

PR #91, opened 2019-05-10
remove all references to title and url [2/3]

jansenk:
change models, tests, serializers, and factories to remove title and url


reillz10 (approved):
👍LGTM

PR #97, opened 2019-05-13
remove title and url from api definition doc

jansenk:


kdmccormick (changes requested):
We want the program title and URL to still be part of the API, just not part of the Program model, because we can fetch that data from the DiscoveryProgram cache now.

jansenk:
> We want the program title and URL to still be part of the API, just not part of the Program model, because we can fetch that data from the DiscoveryProgram cache now.

Ah, my bad, I misunderstood

PR #99, opened 2019-05-14
Try changing the default celery queue name to 'registrar' to avoid queue name collisions

iloveagent57:
## Description

I noticed in stage logs on splunk that celery tasks are getting routed to notifier workers.  This is no good.  I think what happened is that neither notifier or registrar explicitly declare a queue name, so they are both using the default of "celery".  This PR declares a more specific default queue name for registrar.


schenedx (approved):
👍 Let's at least try it

kdmccormick (approved):
approved, pending we revert this after we confirm whether or not it works, and put it in configuration if it does work.

PR #100, opened 2019-05-15
get title and url from cached discovery program

jansenk:
-save title and marketing_url in dicovery_program
-change serializer for programs to get those fields from cached discovery programs
-add a FakeProgramSerializer for serializing FakePrograms from v1_mock to avoid having to change the whole data model in v1_mock

### DIFF
--- a/registrar/apps/api/serializers.py
+++ b/registrar/apps/api/serializers.py
@@ -20,10 +20,12 @@ class ProgramSerializer(serializers.ModelSerializer):
     Serializer for Programs.
     """
     program_key = serializers.CharField(source='key')
+    program_title = serializers.CharField(source='title')
+    program_url = serializers.URLField(source='url')
 
     class Meta:
         model = Program
-        fields = ('program_key',)
+        fields = ('program_key', 'program_title', 'program_url')
 
 
 class ProgramEnrollmentRequestSerializer(serializers.Serializer):
--- a/registrar/apps/enrollments/data.py
+++ b/registrar/apps/enrollments/data.py
@@ -42,6 +42,8 @@ class DiscoveryProgram(object):
         * version (int)
         * loaded (datetime): When data was loaded from Course Discovery
         * uuid (str): Program UUID-4
+        * title (str): Program title
+        * url (str): Program marketing-url
         * active_curriculum_uuid (str): UUID-4 of active curriculum.
         * course_runs (list[DiscoveryCourseRun]):
             Flattened list of all course runs in program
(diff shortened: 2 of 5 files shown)


registrar/apps/api/serializers.py
```diff
@@ -20,10 +21,32 @@ class ProgramSerializer(serializers.ModelSerializer):
...
+    program_title = serializers.SerializerMethodField()
+    program_url = serializers.SerializerMethodField()
 
     class Meta:
         model = Program
-        fields = ('program_key',)
+        fields = ('program_key', 'program_title', 'program_url')
+
+    def _get_cached_field(self, program, field):
+        discovery_program = DiscoveryProgram.get(program.discovery_uuid)
```
iloveagent57: IMO, it would be more useful to define these as properties (called `title` and `url`) on the `Program` model, so that they're useable in the same way from contexts that aren't related to data serialization.  I think you could leave the implementation exactly the same, and just include a comment in the docstrings for each that the data is being read from the cache, or from a REST API call to discovery if there's a cache miss.

The general idea is that there should be one common interface through which data about a Program is initialized and stored.  A user of the Program model shouldn't have to worry separately about fetching data from a service or cache, and they probably don't want to serialize an object just to figure out the title/url.

registrar/apps/api/serializers.py
```diff
@@ -20,10 +21,32 @@ class ProgramSerializer(serializers.ModelSerializer):
...
+        return val
+
+    def get_program_title(self, program):
+        return self._get_cached_field(program, 'title')
+
+    def get_program_url(self, program):
+        return self._get_cached_field(program, 'url')
+
+
+class FakeProgramSerializer(ProgramSerializer):
```
iloveagent57: ...and I think you'd be able to get rid of this class if you do the above.

registrar/apps/enrollments/data.py
```diff
@@ -51,10 +53,12 @@ class DiscoveryProgram(object):
     # so that all old entries will be ignored.
     class_version = 1
 
-    def __init__(self, version, loaded, uuid, active_curriculum_uuid, course_runs):
+    def __init__(self, version, loaded, uuid, title, url, active_curriculum_uuid, course_runs):
```
iloveagent57: This is getting to be a pretty long set of params - maybe we could use **kwargs and `setattr(self, key, value)` inside the body of this __init__?

iloveagent57 (commented):
This looks good in general, I'm suggesting one lightweight change that I think would be a decent improvement (although I'm of course open to what you or others have to say on the matter).

registrar/apps/api/serializers.py
jansenk: That makes sense. I didn't think of using these fields elsewhere, I was thinking that we just were exposing them on this view. But it makes sense

iloveagent57 (approved):

PR #101, opened 2019-05-20
Increase code coverage

jansenk:
The ticket was for test_views but I thought I'd just get as much done as I could.

-removed a bunch from mixins, we weren't using the serializers like we thought, and also the stuff we had written for api-docs auth was completely unused
-wrote tests for accessing the api-docs
-for the sake of coverage, but also because it made sense to me, made the fluch-cache endpoint return 404 if unauthorized
-wrote tests for task exceptional states
-wrote tests for some uncovered errors
-pragma no cover'ed a bunch

### DIFF
--- a/registrar/apps/api/tests/test_views.py
+++ b/registrar/apps/api/tests/test_views.py
@@ -1,0 +1,73 @@
+""" Tests for non-api views """
+import json
+
+from guardian.shortcuts import assign_perm
+from rest_framework.test import APITestCase
+
+from registrar.apps.core import permissions as perms
+from registrar.apps.core.tests.factories import (
+    OrganizationFactory,
+    OrganizationGroupFactory,
+    UserFactory,
+    USER_PASSWORD
+)
+from registrar.apps.enrollments.tests.factories import ProgramFactory
+
+
+class APIDocViewTest(APITestCase):
+    """ Tests for accessing api-docs """
+    path = '/api-docs/'
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.admin_user = UserFactory()
+        assign_perm(perms.ORGANIZATION_WRITE_ENROLLMENTS, cls.admin_user)
+
+        cls.org = OrganizationFactory(name='Test Organization')
+        cls.program = ProgramFactory(
+            managing_organization=cls.org,
+        )
+        cls.org_admin_group = OrganizationGroupFactory(
+            organization=cls.org,
+            role=perms.OrganizationReadWriteEnrollmentsRole.name
+        )
+        cls.org_admin = UserFactory()
+        cls.org_admin.groups.add(cls.org_admin_group)  # pylint: disable=no-member
+
+        cls.no_perms_user = UserFactory()
+
+    def test_access_apidocs_edxadmin(self):
+        self.access_apidocs(self.admin_user, True)
+
+    def test_access_apidocs_orgadmin(self):
+        self.access_apidocs(self.org_admin, True)
+
+    def test_access_apidocs_noperms(self):
+        self.access_apidocs(self.no_perms_user, True)
+
+    def test_unauthenticated(self):
+        self.access_apidocs(None, False)
+
+    def access_apidocs(self, user, expected_paths):
+        """ Helper method for accessing apidocs and making assertions """
+        if user:
+            self.client.login(username=user.username, password=USER_PASSWORD)
+        else:
+            self.client.logout()
+        response = self.client.get(self.path)
+        self.assertEqual(200, response.status_code)
+        self.assert_paths(response, expected_paths)
+
+    def assert_paths(self, response, expected_paths):
+        """ Assert the presence of 'paths' in the response context """
+        spec = None
+        for context in response.context:  # pragma: no branch
+            if 'spec' in context:  # pragma: no branch
+                spec = context['spec']
+                break
+        self.assertIsNotNone(spec)
+        spec = json.loads(spec)
+        paths = spec.get('paths')
+        self.assertIsNotNone(paths)
+        self.assertEqual(expected_paths, bool(paths))
--- a/registrar/apps/api/v1/mixins.py
+++ b/registrar/apps/api/v1/mixins.py
@@ -8,9 +8,7 @@ from django.core.exceptions import (
     PermissionDenied,
 )
 from django.http import Http404
-from django.urls import resolve
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
-from guardian.shortcuts import get_objects_for_user
 from rest_framework.authentication import SessionAuthentication
 from rest_framework.exceptions import NotAuthenticated, ValidationError
 from rest_framework.response import Response
@@ -19,7 +17,7 @@ from rest_framework.status import HTTP_202_ACCEPTED
 from registrar.apps.api import exceptions
 from registrar.apps.api.constants import ENROLLMENT_WRITE_MAX_SIZE
 from registrar.apps.api.mixins import TrackViewMixin
-from registrar.apps.api.serializers import JobAcceptanceSerializer, ProgramEnrollmentRequestSerializer
+from registrar.apps.api.serializers import JobAcceptanceSerializer
 from registrar.apps.api.utils import build_absolute_api_url
 from registrar.apps.enrollments.models import Program
 from registrar.apps.core import permissions as perms
@@ -47,7 +45,7 @@ class AuthMixin(TrackViewMixin):
 
         If None, permissions will be checked globally.
         """
-        return None
+        return None  # pragma: no cover
 
     def get_permission_required(self, _request):
         """
@@ -70,10 +68,6 @@ class AuthMixin(TrackViewMixin):
 
         Overrides APIView.check_permissions.
         """
-        if resolve(request.path_info).url_name == 'api-docs':
-            self.check_doc_permissions(request)
-            return
-
         if not request.user.is_authenticated:
             raise NotAuthenticated()
 
@@ -86,7 +80,7 @@ class AuthMixin(TrackViewMixin):
             required = [required]
         elif isinstance(required, Iterable):
             required = list(required)
-        else:
+        else:  # pragma: no cover
             raise ImproperlyConfigured(
                 'get_permission_required must return string or iterable; ' +
                 'returned {}'.format(required)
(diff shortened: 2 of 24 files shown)


jansenk:
The tests accessing the api-docs are not reflecting the behavior I would expect. no matter which user i make the request as, the code [here](https://github.com/edx/registrar/blob/911accafeb07c6e90db380f330b7fd67cf7ef088/registrar/api_renderer.py#L22) says the user is unauthenticated and removes the path info. I didn't worry too much because my goal was just coverage but if anyone sees what i'm doing wrong here please let me know

registrar/apps/api/v1/mixins.py
```diff
@@ -255,7 +226,7 @@ def add_tracking_data_from_lms_response(self, response):
         """
         Given an API response from the LMS, add necessary tracking data.
         """
-        if response.status_code == 413:
+        if response.status_code == 413:  # pragma: no cover
```
kdmccormick, line 229: Don't we have tests for oversize payloads?

registrar/apps/api/v1_mock/views.py
```diff
@@ -328,7 +328,7 @@ class MockProgramEnrollmentView(MockProgramSpecificViewMixin, EchoStatusesMixin,
     def get_serializer_class(self):
         if self.request.method == 'POST':
             return ProgramEnrollmentRequestSerializer
-        elif self.request.method == 'PATCH':
+        elif self.request.method == 'PATCH':  # pragma: no branch
```
kdmccormick, line 331: Why is the POST branch getting hit but not the PATCH?

schenedx (approved):
What is the code cov increased to for this PR?
The changes are all reasonable to me. Thank you for add more tests.

registrar/apps/api/tests/test_views.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def test_unauthenticated(self):  # overridden because it won't cause a 401
+        self.access_apidocs(None, False)
+
+    def access_apidocs(self, user, expected_paths):
+        response = self.get(self.path, user)
+        self.assertEqual(200, response.status_code)
+        self.assert_paths(response, expected_paths)
+
+    def assert_paths(self, response, expected_paths):
```
iloveagent57, line 62: Is there ever a case where expected_paths is True?

registrar/apps/core/models.py
```diff
@@ -106,12 +106,12 @@ def role_object(self):
...
             if self.role == role.name:
                 return role
-        return None
+        return None  # pragma: no cover
 
     # pylint: disable=arguments-differ
     def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
-        if self._initial_organization:
+        if self._initial_organization:  # pragma: no branch
```
kdmccormick, line 114: I would expect [this](https://github.com/edx/registrar/blob/master/registrar/apps/core/tests/test_models.py#L111) test to cover this branch

registrar/apps/api/v1/mixins.py
iloveagent57, line 229: I'm surprised you had to `no cover` this line.

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -586,6 +586,26 @@ def test_backend_multi_status_response(self):
...
+    @responses.activate
+    def test_backend_server_error(self):
+        self.mock_enrollments_response(self.method, 'Internal Server Error', response_code=500)
+
+        req_data = [
+            self.student_enrollment('active', '001'),
+            self.student_enrollment('active', '002'),
+            self.student_enrollment('inactive', '003'),
+        ]
+        exception_occurred = False
```
iloveagent57: You probably want to use this instead: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex

kdmccormick (changes requested):
Some questions about some of the new no-covers/no-branches.

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -1074,6 +1099,25 @@ def test_backend_multi_status_response(self):
...
+    @responses.activate
+    def test_backend_server_error(self):
+        self.mock_course_enrollments_response(self.method, 'Internal Server Error', response_code=500)
+
+        req_data = [
+            self.student_course_enrollment('active', '001'),
+            self.student_course_enrollment('active', '002'),
+            self.student_course_enrollment('inactive', '003'),
+        ]
+        exception_occurred = False
```
iloveagent57: Same thing with `self.assertRaisesRegex`.

registrar/apps/api/v1/mixins.py
jansenk: We do, but not for the 413 response from the LMS. Unless there's somehow a mismatch in LMS and Registrar limits, we should never get here, but i thought we may as well keep it.

registrar/apps/api/v1_mock/views.py
iloveagent57, line 331: Hmm, do we not have tests for this branch?  Surely there are tests that make PATCH requests to this endpoint.

registrar/apps/core/models.py
iloveagent57, line 114: hmm, we may actually want tests for this condition (doesn't have to be in this PR).

registrar/apps/core/signals.py
```diff
@@ -14,7 +14,7 @@ def handle_user_post_save(sender, **kwargs):  # pylint: disable=unused-argument
     OrganizationGroup and delete the pending record.
     """
     user_instance = kwargs.get("instance")
-    if user_instance is None:
+    if user_instance is None:  # pragma: no cover
```
iloveagent57: I think we should write a test function for this condition.

registrar/apps/core/utils.py
```diff
@@ -52,7 +52,7 @@ def serialize_to_csv(items, field_names, include_headers=False):
     """
     outfile = StringIO()
     writer = csv.DictWriter(outfile, fieldnames=field_names)
-    if include_headers:
+    if include_headers:  # pragma: no cover
```
iloveagent57: lets just have the test for this call the function with `include_headers=True`.

registrar/apps/enrollments/data.py
```diff
@@ -233,7 +233,7 @@ def _get_all_paginated_results(url, client=None):
 
     Repeatedly performs request on 'next' URL until 'next' is null.
     """
-    if not client:
+    if not client:  # pragma: no branch
```
iloveagent57, line 236: We probably want a test for this.

registrar/apps/api/v1_mock/views.py
jansenk: the PATCH branch is being hit, line 331 is still covered. This line got a partial because "line 331 is never false"

iloveagent57 (commented):
Thanks for doing this!  I think there are a few things that you `no covered` that actually deserve their own tests.  Generally looks good, though.

registrar/apps/api/v1/mixins.py
jansenk: same answer as kyle's question here: "We do [have similar tests], but not for the 413 response from the LMS. Unless there's somehow a mismatch in LMS and Registrar limits, we should never get here, but i thought we may as well keep it."

registrar/apps/api/v1_mock/views.py
jansenk: the PATCH branch is being hit, line 331 is still covered. This line got a partial because "line 331 is never false"

registrar/apps/core/models.py
jansenk: this is being hit, the complaint coverage has is that this if statement is never false, which is why it's `no branch`'ed rather than `no cover`'ed

kdmccormick: We definitely do want to test this, and [I thought we did](https://github.com/edx/registrar/blob/master/registrar/apps/core/tests/test_models.py#L111). Maybe the test isn't executing the way I expect it to...

registrar/apps/api/tests/test_views.py
jansenk: no, and i'm not sure why. from my comment above: 

> The tests accessing the api-docs are not reflecting the behavior I would expect. no matter which user i make the request as, the code [here](https://github.com/edx/registrar/blob/911accafeb07c6e90db380f330b7fd67cf7ef088/registrar/api_renderer.py#L22) says the user is unauthenticated and removes the path info. I didn't worry too much because my goal was just coverage but if anyone sees what i'm doing wrong here please let me know

registrar/apps/core/models.py
jansenk: this is being hit, the complaint coverage has is that this if statement is never false, which is why it's no branch'ed rather than no cover'ed

registrar/apps/core/utils.py
jansenk: sounds good, didn't know if we wanted to do that and it's currently inaccessible any other way than calling directly

registrar/apps/enrollments/data.py
jansenk: same as other comment above, the body of the statement is being hit, it's just never false

registrar/apps/api/v1_mock/views.py
kdmccormick: Ah, gotcha.

registrar/apps/core/models.py
kdmccormick: Gotcha! TIL the difference between `no cover` and `no branch`

jansenk:
> What is the code cov increased to for this PR?

99% !

registrar/apps/api/v1/mixins.py
kdmccormick: Right; makes sense to me.

registrar/apps/api/tests/test_views.py
kdmccormick: Let's try to figure this out; I'd rather leave coverage lower than merge tests that make unexpected assertions.

registrar/apps/core/signals.py
jansenk: how?

iloveagent57: Manually call the function without an `instance` kwarg, assert that it logs an error and returns null.

registrar/apps/api/tests/test_views.py
kdmccormick: @jansenk alright I think I got it -- the `render_yaml_spec` view function, unlike all our other endpoints, does *not* accept JWT tokens, which is what all our tests rely on. I think that's OK because the api-docs path is a GUI, not a REST endpoint.

For the tests, instead of using `AuthRequestMixin` with the special JWT-authenticated `self.get/post/patch` methods, I would just use something along the lines of:
```python
self.user = UserFactory(username='abc')
self.user.set_password('123')
self.user.save()
...
self.client.login(username=self.user.username, password='123')
self.client.get('/api-docs')
```

registrar/apps/core/signals.py
iloveagent57: OR!  You can delete these three lines of code.  That's a perfectly acceptable solution to me.

jansenk: Let's go with that one :)

registrar/apps/api/tests/test_views.py
jansenk: Thanks! I'll give it a shot

jansenk: That did it!

kdmccormick (approved):
Great work! Feels good to have 99% coverage.

PR #108, opened 2019-05-28
Fix media & task settings

kdmccormick:
@edx/masters-neem 
[EDUCATOR-4301](https://openedx.atlassian.net/browse/EDUCATOR-4301)

Registrar's config sets `MEDIA_STORAGE_BACKEND`, but we weren't actually loading the settings within that dict into Registrar settings. See https://github.com/edx/configuration/pull/3307#discussion_r74458655 for some historical context.

Two questions for @bderusha:
 1) Is it fine to set CELERY_DEFAULT_QUEUE and CELERY_DEFAULT_ROUTING_KEY here, or should that be done in configuration?
 2) Is it fine to switch from `yaml.load` to `yaml.safe_load`, or do we expect the YAMLs generated by our configuration pipeline to have advanced features that `safe_load` would choke on?

### DIFF
--- a/registrar/settings/production.py
+++ b/registrar/settings/production.py
@@ -12,13 +12,16 @@ ALLOWED_HOSTS = ['*']
 
 LOGGING = get_logger_config()
 
+# This may be overridden by the YAML in REGISTRAR_CFG, but it should be here as a default.
+MEDIA_STORAGE_BACKEND = {}
+
 # Keep track of the names of settings that represent dicts. Instead of overriding the values in base.py,
 # the values read from disk should UPDATE the pre-configured dicts.
 DICT_UPDATE_KEYS = ('JWT_AUTH',)
 
 CONFIG_FILE = get_env_setting('REGISTRAR_CFG')
 with open(CONFIG_FILE, encoding='utf-8') as f:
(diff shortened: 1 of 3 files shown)


kdmccormick:
Hey @bderusha , two questions:
1) Is it fine to set `CELERY_DEFAULT_QUEUE` and `CELERY_DEFAULT_ROUTING_KEY` here, or should that be done in configuration?
2) Is it fine to switch from `yaml.load` to `yaml.safe_load`, or do we expect the YAMLs generated by our configuration pipeline to have advanced features that `safe_load` would choke on?

registrar/settings/production.py
```diff
@@ -12,13 +12,16 @@
 
 LOGGING = get_logger_config()
 
+# This may be overridden by the YAML in DISCOVERY_CFG, but it should be here as a default.
```
iloveagent57: s/DISCOVERY_CFG/REGISTRAR_CFG/

bderusha:
Not sure about the `safe_load` but it will blow up nice and loud if we try and it doesn't work.  Seems reasonable to keep the default config there.  :shipit:

bderusha (approved):

PR #114, opened 2019-05-30
add function to check if there is an in-progress job for a program

jansenk:
add function to check if there is an in-progress job for a program

### DIFF
--- a/registrar/apps/enrollments/utils.py
+++ b/registrar/apps/enrollments/utils.py
@@ -1,0 +1,37 @@
+""" Utilities for the enrollments app """
+from user_tasks.models import UserTaskStatus
+
+
+USER_TASK_STATUS_PROCESSING_STATES = [
+    UserTaskStatus.PENDING,
+    UserTaskStatus.IN_PROGRESS,
+    UserTaskStatus.RETRYING,
+]
+
+
+def build_enrollment_job_status_name(program_key, task_name):
+    """
+    Build the UserTaskStatus.name for the given task and program
+
+    Arguments:
+        program_key (str): program key for the program we're writing enrollments
+        task_name (str): the name of the task that is being executed
+    """
+    return "{}:{}".format(program_key, task_name)
+
+
+def is_enrollment_job_processing(program_key):
+    """
+    Returns whether or not a bulk enrollment job for a particular program
+    is currently processing (in progress, pending, or retrying)
+
+    Used to ensure only one bulk task can be run at a time for a program.
+
+    Arguments:
+        program_key (str): program key for the program we're checking
+    """
+    program_prefix = build_enrollment_job_status_name(program_key, '')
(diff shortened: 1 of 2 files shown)


registrar/apps/core/jobs.py
```diff
@@ -163,3 +163,31 @@ def _affirm_job_in_progress(job_id, task_status):
             'Job can only be marked as Succeeded from state In Progress' +
             '(job_id = {})'.format(job_id)
         )
+
+
+def build_enrollment_job_status_name(program_key, task_name):
+    """
+    Buld the UserTaskStatus.name for the given task and program
```
iloveagent57: Nit "build"

iloveagent57 (commented):
This looks good, but how ya' gonna keep that coverage at 99+% :)

kdmccormick (changes requested):
The code itself looks great! A couple suggestions though --
1) Like Alex mentioned, tests would be good -- `is_enrollment_job_in_progress` assumes certain properties of the implementation of `build_enrollment_job_status_name` (specifically, that prefixing will work), so we should make sure that changing `build_enrollment_job_status_name` in an incompatible way will break tests
2) `core/jobs.py` is meant to be for the infrastructure that makes jobs work with django-user-tasks. I'd put enrollment-specific stuff like this in `enrollments/utils.py`

registrar/apps/enrollments/utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+    Used to ensure only one bulk task can be run at a time for a program.
+
+    Arguments:
+        program_key (str): program key for the program we're checking
+    """
+    program_prefix = build_enrollment_job_status_name(program_key, '')
+    return UserTaskStatus.objects.filter(
+        name__startswith=program_prefix,
+        state=UserTaskStatus.IN_PROGRESS,
```
iloveagent57: we might actually want to check if the state is any of (IN_PROGRESS, PENDING, RETRYING).  If it's either PENDING or RETRYING, it means that there is a task for this program that might change to IN_PROGRESS in the future, and we don't know if it will run before or after the new one a user wants to enqueue.
@jansenk @kdmccormick what do you guys think?

iloveagent57: Also, thanks for writing those tests, it's what prompted my question.

zacharis278 (approved):

registrar/apps/enrollments/utils.py
kdmccormick: @iloveagent57 @jansenk I agree with checking the two other statuses

jansenk: yeah that makes sense. I was thinking of "in progress" in a more abstract sense, but I think those statuses are semantically "in progress"

iloveagent57 (approved):

PR #117, opened 2019-05-31
Fix bug where curriculum UUID breaks enrollment get

kdmccormick:
@edx/masters-neem 
https://openedx.atlassian.net/browse/EDUCATOR-4391

Note: I changed the CSV-writing tests to be about fruit instead of enrollments, because I don't want to give the impression that anything about enrollments is actually being tested there.


iloveagent57 (approved):
👍

PR #118, opened 2019-05-31
Define tasks to write program enrollments from S3 files, et al.

kdmccormick:
@edx/masters-neem 
[EDUCATOR-4377](https://openedx.atlassian.net/browse/EDUCATOR-4377)

This PR:
* Defines a task for batched writing of program enrollments from LMS. It does not define course run enrollment writing, but that should be fairly straightforward to implement, as most of the plumbing (and test scaffolding) is already there.
* Changes how Registrar handles certain cases of bad input -- see https://github.com/edx/registrar/pull/123#issuecomment-499551449.
* Incorporates a commit for @jansenk 's change to improve the Program display on the admin panel.
* Reconciles the return schemas of `/api/v1/jobs/` and `/api/v1/jobs/<job_id>`


registrar/apps/enrollments/data.py
```diff
@@ -227,31 +217,129 @@ def get_course_run_enrollments(program_uuid, course_key, client=None):
...
 
-    try:
-        return _make_request(method, url, client, json=enrollments)
-    except HTTPError as e:
-        response = e.response
-        if response.status_code in [404, 422]:
-            return response
-        raise
+def _lms_program_enrollment_url(program_uuid):
+    """ TODO docstring """
```
zacharis278: 👋 assume you wanted to fill these in

zacharis278 (approved):
a lot here but it looks good to me

registrar/apps/enrollments/data.py
kdmccormick: yes, thank you!

PR #119, opened 2019-06-03
move defaults to registrar

syed-awais-ali:
## Description

### DIFF
--- a/registrar/settings/base.py
+++ b/registrar/settings/base.py
@@ -181,7 +181,7 @@ TEMPLATES = [
 # Detailed information at: https://docs.djangoproject.com/en/dev/ref/settings/
 SESSION_COOKIE_NAME = 'registrar_sessionid'
 CSRF_COOKIE_NAME = 'registrar_csrftoken'
-LANGUAGE_COOKIE_NAME = 'registrar_language'
+LANGUAGE_COOKIE_NAME = 'openedx-language-preference'
 # END COOKIE CONFIGURATION
 
 # AUTHENTICATION CONFIGURATION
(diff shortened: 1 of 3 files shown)


kdmccormick (changes requested):
Reviewing now

registrar/settings/base.py
```diff
@@ -260,3 +261,19 @@
...
 )
+
+CERTIFICATE_LANGUAGES = {
+    'en': 'English',
+    'es_419': 'Spanish',
+}
+
+CSRF_COOKIE_SECURE = False
+EXTRA_APPS = []
+MEDIA_STORAGE_BACKEND = {
```
kdmccormick: `MEDIA_STORAGE_BACKEND` isn't an actual setting; it's an edX-configuration-specific dict we unpack [only in production.py](https://github.com/edx/registrar/blob/master/registrar/settings/production.py#L35), so this is a no-op.

We already set the media storage defaults [in base.py here](https://github.com/edx/registrar/blob/master/registrar/settings/base.py#L134-L137), so you can safely delete this dict :)

kdmccormick (approved):
Pending the once change I requested, this looks good to me.

registrar/settings/base.py
```diff
@@ -245,7 +246,7 @@
 SEGMENT_KEY = None
 
 # Publicly-exposed base URLs for service and API
-API_ROOT = 'http://replace-me/api'
+API_ROOT = 'http://localhost:18734/api'
```
adzuci: I don't have enough context to have a strong oppinion, but I'm curious if we should default to a running local configuration or if it's ok to have a few `<replaceme>`s and allow the community to plug in what's needed?

adzuci (approved):

registrar/settings/base.py
kdmccormick: ~Having `localhost:port` baked in here will make things more likely to "just work" if someone spins up Registrar in devstack for the first time, and I think there's value in that. However, you could definitely argue that such devstack-specific configuration should instead be put in devstack, which would reap the same benefits while also avoiding mixing together config from different parts of our ecosystem. In the end, though, I'm so confused about our overall configuration strategy that I don't have a strong opinion either.~

Forgot we had `devstack.py`, see next comment.

kdmccormick: I think the best compromise at the moment would be putting devstack-specific stuff in Registrar's `devstack.py` and putting `replace-me`s in `base.py` .

( @syed-awais-ali )

kdmccormick (approved):
@syed-awais-ali  Looks good -- the only change I request is that you keep `API_ROOT = http://replace-me/api` in `base.py` and override as:
* `API_ROOT = 'http://localhost:18734/api'` in `devstack.py`
* `API_ROOT = 'http://localhost:8000/api'` in `local.py`

syed-awais-ali:
@kdmccormick Addressed the changes

syed-awais-ali:
We already have `API_ROOT = 'http://localhost:18734/api'` in local.py and not `API_ROOT = 'http://localhost:8000/api'`
Added `API_ROOT = 'http://localhost:18734/api'` to the devstack.py.
@kdmccormick please can you again take a look at this. Thanks

kdmccormick:
@syed-awais-ali  Thanks for adding `API_ROOT = 'http://localhost:18734/api'` to devstack.py.

The `API_ROOT = 'http://localhost:18734/api'` in local.py is a mistake; please change it to `API_ROOT = 'http://localhost:8000/api'`, and then you have my go-ahead to squash and merge!

kdmccormick (approved):
Thanks @syed-awais-ali ! Go ahead and sqaush & merge

PR #121, opened 2019-06-04
Jkantor/handle 404

jansenk:
_make_request will still raise an exception on a 404 from LMS, but it will now be caught by write_program_course_enrollments and write_program_enrollments

I added test cases to make the behavior more explicit


iloveagent57 (approved):
LGTM

kdmccormick (commented):
Generally looks good, but I'd recommend also adding a short test case to [the enrollment write API view tests](https://github.com/edx/registrar/blob/master/registrar/apps/api/v1/tests/test_views.py#L728) along the lines of `test_program_not_found_in_lms`. Before this change we return 500 (which is what the bug ticket reports); not sure if this should return 404, 400, or something else.

kdmccormick (approved):
LGTM! Just name the commit something more descriptive :)

PR #122, opened 2019-06-04
Set celery exchange, rename default queue

jdmulloy:
We need to also set the celery exchange setting. I also made the default queue name 'registrar.default' in case we ever want more queues. It's fine to leave it as just registrar if we think the chance of other queues is low. I'd like the have all queues include the app name in them, so if a new queue is ever needed it should be registrar,foobar instead of just foobar so we know which IDA it goes with.

### DIFF
--- a/registrar/settings/base.py
+++ b/registrar/settings/base.py
@@ -109,8 +109,9 @@ CELERYD_HIJACK_ROOT_LOGGER = False
 # only registrar workers should receive registrar tasks.
 # explicitly define this to avoid name collisions with other services
 # using the same broker and the standard default queue name of "celery"
-CELERY_DEFAULT_QUEUE = 'registrar'
+CELERY_DEFAULT_EXCHANGE = 'registrar'
 CELERY_DEFAULT_ROUTING_KEY = 'registrar'
+CELERY_DEFAULT_QUEUE = 'registrar.default'
 
 ############################# END CELERY #################################3
 
--- a/registrar/settings/base.py
+++ b/registrar/settings/base.py
@@ -109,9 +109,9 @@ CELERYD_HIJACK_ROOT_LOGGER = False
 # only registrar workers should receive registrar tasks.
 # explicitly define this to avoid name collisions with other services
 # using the same broker and the standard default queue name of "celery"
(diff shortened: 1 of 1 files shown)


kdmccormick (commented):
Sounds good, but I think we need to update https://github.com/edx/configuration/blob/master/playbooks/roles/registrar/defaults/main.yml#L142 too.

The right solution, of course, would be to set Registrar's `CELERY_DEFAULT_QUEUE` value based off the linked queue name setting...

Also, TBH, I don't know the difference between queue, routing key, and exchange, but I'll trust your knowledge that these values are sensible.

iloveagent57:
https://docs.celeryproject.org/en/latest/userguide/routing.html#exchanges-queues-and-routing-keys

Yeah, we'll have to update the configuration code, too.  That code does stuff like telling the workers which queue they should listen to when supervisor spins them up.

iloveagent57 (approved):
Tagged you guys on this corresponding PR: https://github.com/edx/configuration/pull/5172

kdmccormick:
@jdmulloy @iloveagent57 

how about:
```python
CELERY_DEFAULT_EXCHANGE = os.environ.get('CELERY_DEFAULT_EXCHANGE', 'registrar')
CELERY_DEFAULT_ROUTING_KEY = os.environ.get('CELERY_DEFAULT_ROUTING_KEY', 'registrar')
CELERY_DEFAULT_QUEUE = os.environ.get('CELERY_DEFAULT_QUEUE', 'registrar.default')
```

and then in https://github.com/edx/configuration/pull/5172:
```yaml
registrar_celery_default_queue: registrar.default
registrar_environment:
  ....
  CELERY_DEFAULT_EXCHANGE: registrar
  CELERY_DEFAULT_ROUTING_KEY: registrar
  CELERY_DEFAULT_QUEUE: "{{ registrar_celery_default_queue }}"
...
...
REGISTRAR_CELERY_WORKERS:
  - queue: "{{ registrar_celery_default_queue }}"
    ...

iloveagent57:
@kdmccormick I like your suggestion, commit added to implement it.

kdmccormick (approved):
Awesome :ship:

PR #123, opened 2019-06-06
add test

jansenk:
I found this while looking into EDUCATOR-4390. If we don't provide a student_key or status to the registrar, it will insert the fields as None when it makes its request to the LMS. I'm not sure that we really need to do anything here, since EDUCATOR-4390 is fixed by making the LMS endpoint resilient to this behavior. I could see us adding checks against this, so I just thought I'd add the test for completeness' sake.

Also I modified the Program django admin in the course of debugging and I think it's better this way.

### DIFF
--- a/registrar/apps/api/v1/tests/test_views.py
+++ b/registrar/apps/api/v1/tests/test_views.py
@@ -660,6 +660,7 @@ class ProgramCourseListViewTests(RegistrarAPITestCase, AuthRequestMixin):
         self.assertEqual(response.status_code, 404)
 
 
+@ddt.ddt
 class ProgramEnrollmentWriteMixin(object):
     """ Test write requests to the /api/v1/programs/{program_key}/enrollments endpoint """
     path = 'programs/masters-in-english/enrollments'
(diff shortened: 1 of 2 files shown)


kdmccormick:
Interesting behavior; good catch. In https://github.com/edx/registrar/pull/118 I refactored Registrar to return 400 if the `student_key` is missing or not a string, and blindly pass the data to LMS in any other case.

The reason I needed to check for `student_key` in Registrar is because we now need to check for duplicate keys in Registrar to enable the batched writing of enrollments from uploaded CSVs. To do duplicate-checking, I needed to make sure the student keys were there.

As for validating the presence of the `status` field, I could:
1) Do what I have in my PR now, which is ignore it and let LMS handle i.
2) Do what we do on `master`, which is pass `None` to LMS if the status is missing, which is what you test for here.
3) Registrar raises 400 if any statuses are missing.

Let me know what you think! Oh, and the admin stuff looks good :+1: 

Edit - for reference, here is where I do the `student_key` validation in my PR: https://github.com/edx/registrar/pull/118/files#diff-5307a5ded526aa0d82da93ff2e120d19R249

jansenk:
I think I'd vote for option 3.
I know the original idea was to ignore as much as we could and let the LMS handle it, but if you're already adding a check for the presence of the student_key, might as well toss in a check for the status for good measure.
If you do that, the behavior I'm testing here is basically gone (in a good way) so I think I'll just close this. We can change the django admin some other time.

kdmccormick:
@jansenk  Cool, I integrated the third option into https://github.com/edx/registrar/pull/118, plus a couple DDT cases confirming it. Also, I added a commit for the program admin change.

PR #125, opened 2019-06-10
Jkantor/external course key

jansenk:
- Added methods on DiscoveryProgram for getting internal course key and external course key
- For course-related endpoints, do conversion to internal course key right away.

The ticket mentions that conversion internal -> external was needed for user tasks, but I don't actually see anywhere where we include the key in task results. The functionality is there and tested, but not used yet


zacharis278 (approved):
looks good 👍

iloveagent57 (approved):
Nice!

PR #127, opened 2019-06-11
add job listing endpoint

jansenk:
added new endpoint /api/v1/jobs/ that lists the requesting user's currently processing jobs

### DIFF
--- a/registrar/apps/api/v1/views.py
+++ b/registrar/apps/api/v1/views.py
@@ -21,6 +21,7 @@ from registrar.apps.api.serializers import (
     CourseRunSerializer,
     JobStatusSerializer,
     ProgramSerializer,
+    UserTaskStatusSerializer,
 )
 from registrar.apps.api.v1.mixins import (
     AuthMixin,
@@ -42,6 +43,7 @@ from registrar.apps.enrollments.tasks import (
     list_course_run_enrollments,
     list_program_enrollments,
 )
+from registrar.apps.enrollments.utils import get_processing_jobs_for_user
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


registrar/apps/api/v1/views.py
```diff
@@ -423,3 +425,23 @@ def get_object(self):
...
+
+    Returns:
+     * 200: OK
+     * 401: User is not logged in.
+    """
+
+    authentication_classes = (JwtAuthentication, SessionAuthentication)
+    permission_classes = (IsAuthenticated,)
+    serializer_class = UserTaskStatusSerializer
+    event_method_map = {'GET': 'registrar.v1.list_job_statuses'}
```
iloveagent57, line 444: What does this do?

jansenk: event method map? this was added a little while ago for segment tracking

iloveagent57: Ah, right, thanks!

registrar/apps/enrollments/utils.py
```diff
@@ -35,3 +35,18 @@ def is_enrollment_job_processing(program_key):
...
+
+    Arguments:
+        user (User): User attempting to read job statuses.
+
+    Returns: list of processing UserTaskStatuses for the given user
+    """
+    return UserTaskStatus.objects.filter(
+        user=user,
+        state__in=USER_TASK_STATUS_PROCESSING_STATES
+    ).all()
```
iloveagent57: Curious if you need the `.all()` here.

iloveagent57 (approved):
Looks good!  Have that one outstanding question, but I think you're good to go.

registrar/apps/enrollments/utils.py
jansenk: You definitely don't, but it seemed ... better(?) to return a list rather than a QuerySet .... but this is only really called in getQuerySet .... okay yeah qs it is

PR #128, opened 2019-06-12
change some tests and api.yml

jansenk:


api.yaml
```diff
@@ -43,7 +43,7 @@ course_path_param: &course_path_param
   name: course_id
   required: true
   type: string
-  description: edX course run identifier
+  description: edX course run identifier, or a course run external course key
```
kdmccormick: `, or an external course run key` might read a little better.

api.yaml
```diff
@@ -654,6 +654,8 @@ models:
     properties:
       course_id:
         type: string
+      external_course_key:
```
kdmccormick, line 658: Can you also update the example that uses this model on Line 197?

kdmccormick (approved):
Cool, thanks for doing this. Just two comments, but otherwise looks good.

PR #131, opened 2019-06-18
csv upload endpoints

zacharis278:
### [EDUCATOR-4335](https://openedx.atlassian.net/browse/EDUCATOR-4335)

* endpoints to upload program/course enrollments via csv
* added a 5 MiB upper limit on file size jic

### DIFF
--- a/registrar/apps/api/v1/views.py
+++ b/registrar/apps/api/v1/views.py
@@ -439,7 +439,7 @@ class EnrollmentUploadView(JobInvokerMixin, APIView):
 
         enrollments = []
         reader = csv.DictReader(file_itr)
-        if not reader.fieldnames == self.field_names:
+        if reader.fieldnames != self.field_names:
             raise ValidationError('Invalid csv headers')
 
         for n, row in enumerate(reader, 1):
--- a/registrar/apps/api/v1/views.py
+++ b/registrar/apps/api/v1/views.py
@@ -1,6 +1,9 @@
 """
 The public-facing REST API.
 """
+import csv
+import io
+import json
 import logging
 
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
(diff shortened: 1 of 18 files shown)


zacharis278:
added in a fix for the filestore differences between s3 and local storage
> https://github.com/edx/registrar/pull/131/files#diff-2213b9fd93196ca1860ba79af3cc5fc2R57

registrar/apps/enrollments/tasks.py
```diff
@@ -157,6 +155,13 @@ def write_program_enrollments(
     post_job_success(job_id, results_str, "csv", text=code_str)
 
 
+@shared_task(base=EnrollmentWriteTask, bind=True)
+def write_course_run_enrollments(
+        self, job_id, user_id, json_filepath, program_key
+):
+    pass
```
kdmccormick: I'd change this to fail with message "not implemented" so that we don't accidentally put a program in a locked state.

registrar/apps/api/v1/views.py
```diff
@@ -391,3 +405,77 @@ class JobStatusListView(AuthMixin, TrackViewMixin, ListAPIView):
...
+    Returns:
+     * 202: Upload created
+     * 403: User lack write access at specified program
+     * 400: Validation error, missing or invalid file
+     * 404: Program does not exist
+     * 409: Job already processing for this program
+     * 413: File too large
+    """
+    parser_classes = (MultiPartParser,)
+
```
kdmccormick, line 425: Nit: For "abstract" classes like this, I like throwing something like `task_fn = None  # Override in subclass` in the body to make it clear which members need to get overridden.

kdmccormick (commented):
Looks great! Two smallish comments.

registrar/apps/enrollments/tasks.py
zacharis278: that's a very good point.

registrar/apps/api/v1/views.py
```diff
@@ -391,3 +405,78 @@ class JobStatusListView(AuthMixin, TrackViewMixin, ListAPIView):
...
+     * 202: Upload created
+     * 403: User lack write access at specified program
+     * 400: Validation error, missing or invalid file
+     * 404: Program does not exist
+     * 409: Job already processing for this program
+     * 413: File too large
+    """
+    parser_classes = (MultiPartParser,)
+    task_fn = None  # Override is subclass
+
```
kdmccormick, line 425: Also `field_names`

kdmccormick (changes requested):
Code looks good other than a final nit.

However I get a 500 when I POSTed without a file or a trailing slash:
```
RuntimeError at /api/v1/programs/gtmasters/enrollments/upload

You called this URL via POST, but the URL doesn't end in a slash and you have APPEND_SLASH set. Django can't redirect to the slash URL while maintaining POST data. Change your form to point to localhost:18734/api/v1/programs/gtmasters/enrollments/upload/ (note the trailing slash), or set APPEND_SLASH=False in your Django settings.
```

kdmccormick (approved):
:zap:

PR #134, opened 2019-06-20
add create_organization and create_user management commands

jansenk:
create two new management commands:

`create_organization <org-name> [--group role-name [group-name]]*`
create an organization with the given name, and for each --group, create a group with the given role, and optionally a name (default <org_name>_<role_name>)

`create_user <username> <email> [--superuser] [--staff] [--groups group-name]`
creates a user with the given username and email, optionally marked as superuser and/or staff,  and add them to specified groups

### DIFF
--- a/registrar/apps/core/management/commands/create_organization.py
+++ b/registrar/apps/core/management/commands/create_organization.py
@@ -1,0 +1,81 @@
+""" Management command to create an Organization and associated OrganizationGroups """
+import logging
+import re
+import uuid
+
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+
+from registrar.apps.core.constants import ORGANIZATION_KEY_PATTERN
+from registrar.apps.core.models import Organization, OrganizationGroup
+from registrar.apps.core.permissions import ORGANIZATION_ROLES
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    # pylint: disable=missing-docstring
+
+    help = 'Creates an Organization with the given key, and any specified OrganizationGroups'
+    role_names = [role.name for role in ORGANIZATION_ROLES]
+
+    def add_arguments(self, parser):
+        parser.add_argument('org_key')
+        parser.add_argument(
+            '--group',
+            nargs='+',
+            dest='groups',
+            action='append',
+            default=[],
+            help='Create an OrganizationGroup. Args: role [name] \n acceptable roles: {}'.format(self.role_names)
+        )
+
+    # pylint: disable=arguments-differ
+    @transaction.atomic
+    def handle(self, org_key, groups, *args, **options):
+        groups = self.parse_groups(groups)
+        org = self.create_organization(org_key)
+        for group in groups:
+            self.create_org_group(org, group[0], group[1])
+
+    def parse_groups(self, groups):
+        result = []
+        for group in groups:
+            if len(group) > 2:
+                raise CommandError('--group only accepts one or two arguments')
+            role = group[0]
+            if group[0] not in self.role_names:
+                raise CommandError('first argument to --group must be one of {}'.format(self.role_names))
+            group_name = None
+            if len(group) == 2:
+                group_name = group[1]
+            result.append((role, group_name))
+        return result
+
+    def create_organization(self, org_key):
+        if not re.fullmatch(ORGANIZATION_KEY_PATTERN, org_key):
+            raise CommandError('org_key can only contain alphanumeric characters, dashes, and underscores')
+        try:
+            org = Organization.objects.create(
+                key=org_key,
+                name=org_key,
+                discovery_uuid=uuid.uuid4(),
+            )
+        except Exception as e:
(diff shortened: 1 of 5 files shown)


zacharis278 (approved):
lgtm 👍

jansenk:
Rebase and simplify log messages thanks to #135

kdmccormick:
taking a quick look :eyes:

registrar/apps/core/management/commands/test/test_create_user.py
```diff
@@ -0,0 +1,112 @@
...
+        expected_group_names=None
+    ):
+        """ Assert that the given user has certain values """
+        if expected_username is None:  # pragma: no branch
+            expected_username = self.username
+        if expected_email is None:  # pragma: no branch
+            expected_email = self.email
+        if expected_group_names is None:
+            expected_group_names = []
+        self.assertEqual(user.username, expected_username)
```
kdmccormick: Nit: you can avoid these types of `if`s by doing stuff like `self.assertEqual(user.email, expected_email or self.email)`

registrar/apps/core/management/commands/test/test_create_user.py
```diff
@@ -0,0 +1,112 @@
...
+        self.assertEqual(user.email, expected_email)
+        self.assertEqual(user.is_superuser, is_superuser)
+        self.assertEqual(user.is_staff, is_staff)
+        self.assertEqual(set(group.name for group in user.groups.all()), set(expected_group_names))
+
+    def test_create_user(self):
+        call_command(self.command, self.username, self.email)
+        self.assert_user(User.objects.get(username=self.username))
+
+    @ddt.data(
```
kdmccormick, line 52: :+1: on the clever ddt

registrar/apps/core/management/commands/create_organization.py
```diff
@@ -0,0 +1,77 @@
...
+            group_name = None
+            if len(group) == 2:
+                group_name = group[1]
+            result.append((role, group_name))
+        return result
+
+    def create_organization(self, org_name):
+        try:
+            org = Organization.objects.create(
+                key=org_name,
```
kdmccormick: because we're using `org_name` as the key, consider failing if it contains anything except `[A-Za-z0-9-_]`

registrar/apps/core/management/commands/create_organization.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class Command(BaseCommand):
+    # pylint: disable=missing-docstring
+
+    help = 'Creates an Organization with the given name, and any specified OrganizationGroups'
+    role_names = [role.name for role in ORGANIZATION_ROLES]
+
+    def add_arguments(self, parser):
+        parser.add_argument('org_name')
```
kdmccormick: I'd call this `'org_key'`, because that's what ends up mattering in the API. The fact that it is the org name too is more of a detail.

kdmccormick (commented):
Looks generally great, a few suggestions below. My only real concerns are:
1) Will the created users automatically link up to the corresponding LMS users? I'm trying this now on my devstack, I'll let you know if it works.
2) Do we need a command to create programs too? I know Discovery is going to get populated from prod data, but that won't automatically populate the Registrar `Program` table.

EDIT: Manual testing makes it seem like (1) should be fine; I guess Registrar automatically matches the username in the JWT up with the user in Registrar (or creates a new one if no matching user exists)

kdmccormick:
@jansenk can you also make `email` optional? It looks like the username is what matters with JWT auth, and the email in Regsitrar gets overwritten with the email in the JWT, so we'd be wasting keystrokes by manually entering in emails in to Registrar.

registrar/apps/core/management/commands/test/test_create_user.py
jansenk: it's funny, i'm used to doing things like this in other languages / contexts, but for some reason I never even thought to do it in python

jansenk:
> @jansenk can you also make `email` optional? It looks like the username is what matters with JWT auth, and the email gets overwritten with the email in the JWT, so we'd be wasting keystrokes by manually entering in emails.

Interesting. I knew that the jwt was only looking at the username, but I never would have thought that it actively overrides the email.

registrar/apps/core/management/commands/create_user.py
```diff
@@ -0,0 +1,70 @@
...
+    def get_groups(self, group_names):
+        if not group_names:
+            return []
+        group_set = set(group_names)
+        if len(group_set) != len(group_names):
+            raise CommandError('Duplicate groups not allowed')
+        groups = []
+        for group_name in group_names:
+            try:
+                group = OrganizationGroup.objects.get(name=group_name)
```
kdmccormick: because this seems like a general user-creation command, I'd query the standard auth `Group` table here, which will include all the `OrganizationGroup`s.

kdmccormick (commented):
Other than the comment I left, my only remaining question is the one about programs:

> Do we need a command to create programs too? I know Discovery is going to get populated from prod data, but that won't automatically populate the Registrar Program table.

PR #138, opened 2019-06-25
Better write logging & response validation

kdmccormick:
@edx/masters-neem

### DIFF
--- a/registrar/apps/enrollments/data.py
+++ b/registrar/apps/enrollments/data.py
@@ -1,6 +1,7 @@
 """
 Module for syncing data with external services.
 """
+import json
 import logging
 from collections import namedtuple
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


registrar/apps/enrollments/data.py
```diff
@@ -305,22 +306,40 @@ def key_fn(e):
         for student_key in duplicated_student_keys
     })
     for response in responses:
-        status = response.status_code
-        if status == HTTP_200_OK:
+        unexpected_status = False
```
schenedx: Are these flags only being used for logging purposes?

registrar/apps/enrollments/data.py
```diff
@@ -305,22 +306,40 @@ def key_fn(e):
...
-                "Unexpected status {} from LMS request {} {}. Response body: {}".format(
-                    response.status_code, method, url, response.text
+        try:
+            response_data = response.json()
+        except json.JSONDecodeError:
+            response_data = None
+        if isinstance(response_data, dict):
+            # Only update with student keys that were passed in
+            # so we don't get things like "developer_message" in results.
+            student_data = {
```
schenedx: Even with your comment here, I still fail to understand why do we need to update student_keys this way. Seems to me this is to fix a bug from previous logic.

schenedx (commented):
Can we separate out the logic between data processing and logging? It seems we are making logging smart, but do we need to do that? Can we make logging a bit dump and log for all the status? But just tailor the content of the logging differently based on the status?

registrar/apps/enrollments/data.py
kdmccormick: Yeah, I guess I just don't know where the line between "too smart logging" and "too much logging" is.

Do you think we should just log all response bodies for statuses that are >=400?

registrar/apps/enrollments/data.py
kdmccormick: The way LMS `program_enrollments` is written now, we sometimes get 422s with:
```json
{
    "developer_message": "something something validation error"
}
```
which can cause a Registrar user to get a CSV like:
```
student_key,status
alice,enrolled
bob,pending
charlie,invalid-status
developer_message,something something validation error
```
So we have to do some convoluted key filtering here to avoid that.

This issue was the impetus for that LMS response codes ticket, in hopes that the Registrar side could be a very straightforward:
```python
if status in {200, 207, 422}:
    update results dictionary
else:
   log errors
```

registrar/apps/enrollments/data.py
schenedx: Yeah. Or even all status >200

schenedx (approved):
👍

PR #139, opened 2019-06-25
EDUCATOR-4439 | Implement user task to write course run enrollments from JSON file

iloveagent57:
https://openedx.atlassian.net/browse/EDUCATOR-4439

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ pycodestyle:  # run pycodestyle
 	pycodestyle registrar scripts
 
 pylint:  # run pylint
-	pylint --rcfile=pylintrc registrar scripts
+	pylint --rcfile=pylintrc registrar
 
 yamllint:  # run yamlint
 	yamllint *.yaml
--- a/registrar/apps/enrollments/tasks.py
+++ b/registrar/apps/enrollments/tasks.py
@@ -2,6 +2,7 @@
 This module contains the celery task definitions for the enrollment project.
 """
 import json
+from collections import OrderedDict, namedtuple
 
 from celery import shared_task
 from celery.utils.log import get_task_logger
@@ -13,7 +14,9 @@ from user_tasks.tasks import UserTask
 from registrar.apps.core.constants import UPLOADS_PATH_PREFIX
 from registrar.apps.core.filestore import get_filestore
 from registrar.apps.core.jobs import post_job_failure, post_job_success
+from registrar.apps.core.utils import serialize_to_csv
 from registrar.apps.enrollments import data
+from registrar.apps.enrollments.constants import EnrollmentWriteStatus
 from registrar.apps.enrollments.models import Program
 from registrar.apps.enrollments.serializers import (
     serialize_course_run_enrollments_to_csv,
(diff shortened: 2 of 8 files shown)


registrar/apps/enrollments/tasks.py
```diff
@@ -131,43 +131,57 @@ def write_program_enrollments(
...
     """
-    post_job_failure(  # pragma: no cover
-        job_id,
-        "not implemented",
+    program = _get_program(job_id, program_key)
+    if not program:
+        return
+    requests = _load_enrollment_requests(job_id, program_key, json_filepath)
+    if requests is None:
+        return
```
schenedx, line 182: Should we log something in this condition to say the input file might be empty?

registrar/apps/enrollments/tasks.py
```diff
@@ -131,43 +131,57 @@ def write_program_enrollments(
...
-        # This only happens if no enrollments are given.
-        code_str = "204 No Content"
-    post_job_success(job_id, results_str, "csv", text=code_str)
+    _store_enrollment_task_results(job_id, response_json, any_successes, any_failures)
 
 
 @shared_task(base=EnrollmentWriteTask, bind=True)
 def write_course_run_enrollments(
-        self, job_id, user_id, json_filepath, program_key
+        self, job_id, user_id, json_filepath, program_key, course_key
```
schenedx: I thought the task won't be able to take in a `course_key` because there could be multiple course specified in the course enrollment CSV. Where will the grouping be done?

registrar/apps/enrollments/tasks.py
iloveagent57: That happens downstream, when `_load_enrollment_requests` calls `post_job_failure` in the case that the json file is empty or not parsable.

schenedx (commented):
See my inline questions

registrar/apps/enrollments/tasks.py
iloveagent57: Where do we say that?

Makefile
```diff
@@ -79,7 +79,7 @@ pycodestyle:  # run pycodestyle
 	pycodestyle registrar scripts
 
 pylint:  # run pylint
-	pylint --rcfile=pylintrc registrar scripts
```
schenedx, line 82: We don't care about pylint on scripts anymore?

registrar/apps/enrollments/tasks.py
```diff
@@ -125,49 +128,105 @@ def generate_name(cls, arguments_dict):
         )
 
 
+class EnrollmentWriteStatus(Enum):
```
schenedx: Shall we move these to the constants.py file?

registrar/apps/enrollments/tests/test_tasks.py
```diff
@@ -215,6 +215,22 @@ def tearDown(self):
...
         uploads_filestore.delete(self.json_filepath)
 
+    def mock_write_enrollments(self, any_successes, any_failures):
+        """
+        Creates a mock function that returns results normally returned
+        by an enrollment-writing function.
+        """
+        raise NotImplementedError()  # pragma: no cover
+
+    def test_empty_list_file(self):
```
schenedx, line 225: Nice test case!

schenedx (approved):
All minor feedbacks. It would be great to see a demo of this change

Makefile
kdmccormick: Or Registrar, apparently. Bold move cotton.

iloveagent57: Because the devstack image runs python 3.5 and the travis image is running python 3.6+.  3.6 provides a `random.choices` method, which throws off a `pylint: disable=no-member` pragma that was in a couple of scripts.  Dealing with that felt like more work than it was worth, and it felt sufficiently safe to not run pylint on the non-production code in `scripts/`.

iloveagent57: @kdmccormick see the line of code below this conversation :)

registrar/apps/enrollments/tasks.py
iloveagent57: Maybe?  It's only used in this one file at the moment.

registrar/apps/enrollments/tests/test_tasks.py
iloveagent57: I copied it from another test case!

registrar/apps/enrollments/tasks.py
schenedx: I think aggregate these constants to the constants file make things looks cleaner. Functionally, they don't really mean anything.

PR #142, opened 2019-06-27
Remove refs to amazon-apigateway

iloveagent57:
https://openedx.atlassian.net/browse/EDUCATOR-4442

### DIFF
--- a/api.yaml
+++ b/api.yaml
@@ -11,26 +11,6 @@
 ################################################################################
 swagger: '2.0'
 
-x-amazon-apigateway-integration: &api_gateway_integration
-  type: http_proxy
-  httpMethod: GET
-
-x-amazon-apigateway-integration-job: &api_gateway_integration_job
-  <<: *api_gateway_integration
-  requestParameters:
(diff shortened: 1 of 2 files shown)


api.yaml
```diff
@@ -464,178 +402,6 @@ consumes:
...
-
-paths:
-
-  ### V1 Mock API ###
-  '/v1-mock/jobs/{job_id}/':
-    get:
-      <<: *get_job_status
-      operationId: registrar_v1_mock_get_job_status
-      tags: ['v1 - Mock API']
-      x-amazon-apigateway-integration:
```
zacharis278, line 479: We'll still need the path configuration for the api-docs page to work.  We should just remove these x-amazon-... blocks but keep the rest here

zacharis278 (changes requested):

api.yaml
iloveagent57: 🤦‍♂ That explains why I had to muck with unittests!  Fixed now.

schenedx (approved):
👍

zacharis278 (approved):

PR #143, opened 2019-06-27
Jkantor/external key csv download

jansenk:
Add displayed_course_key optional arg to data. get_course_run_enrollments where it or
course_key are passed into the serializer context and inserted into the enrollment data returned by the LMS

### DIFF
--- a/registrar/apps/api/v1/mixins.py
+++ b/registrar/apps/api/v1/mixins.py
@@ -154,14 +154,14 @@ class CourseSpecificViewMixin(ProgramSpecificViewMixin):
     A mixin for views that operate on or within a specific program course run.
 
     In addition to the functionality provided in `ProgramSpecificViewMixin`,
-    this mixin provides a `internal_course_key` property, which returns the
-    edX course key of the course referred to in the URL (either by the
+    this mixin provides the `internal_course_key` and `external_course_key` properties,
+    which return the edX course key of the course referred to in the URL (either by the
     edX course key itself or by an external course key). It also validates
-    that they course key exists within the program, raising 404 if not.
+    that the course key exists within the program, raising 404 if not.
     """
 
     @cached_property
-    def internal_course_key(self):
+    def course_run(self):
         """
         Raises a 404 if the course run identified by the `course_id` path
         parameter is not part of self.program.
@@ -173,7 +173,15 @@ class CourseSpecificViewMixin(ProgramSpecificViewMixin):
         if not real_course_run:
             self.add_tracking_data(failure='course_not_found')
             raise Http404()
-        return real_course_run.key
+        return real_course_run
+
+    @property
+    def internal_course_key(self):
+        return self.course_run.key
+
+    @property
+    def external_course_key(self):
+        return self.course_run.external_key
 
 
 class JobInvokerMixin(object):
--- a/registrar/apps/enrollments/data.py
+++ b/registrar/apps/enrollments/data.py
@@ -200,13 +200,14 @@ def get_program_enrollments(program_uuid, client=None):
     return serializer.validated_data
 
 
-def get_course_run_enrollments(program_uuid, course_key, client=None):
+def get_course_run_enrollments(program_uuid, internal_course_key, external_course_key=None, client=None):
     """
     Fetches program course run enrollments from the LMS.
 
     Arguments:
         program_uuid (str): UUID-4 string
-        course_key (str): edX course key identifying course run
+        internal_course_key (str): edX course key identifying course run
+        external_course_key (str): optional external course key
 
     Returns: list[dict]
         A list of enrollment dictionaries, validated by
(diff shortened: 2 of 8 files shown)


registrar/apps/enrollments/data.py
```diff
@@ -200,7 +200,7 @@ def get_program_enrollments(program_uuid, client=None):
     return serializer.validated_data
 
 
-def get_course_run_enrollments(program_uuid, course_key, client=None):
+def get_course_run_enrollments(program_uuid, course_key, effective_course_key=None, client=None):
```
schenedx: I am still confused on which one is the `effective_course_key`. Is it the Combined course_key we need to use? If so, could you please update the arguments comments to reflect that?
I also think we should rename `course_key` to something like `internal_course_key` to make it explicit

registrar/apps/enrollments/data.py
```diff
@@ -218,9 +218,10 @@ def get_course_run_enrollments(program_uuid, course_key, client=None):
     """
     url = _lms_course_run_enrollment_url(program_uuid, course_key)
     enrollments = _get_all_paginated_results(url, client)
-    serializer = CourseEnrollmentSerializer(data=enrollments, many=True)
+    context = {'course_key': effective_course_key or course_key}
+    serializer = CourseEnrollmentSerializer(data=enrollments, many=True, context=context)
     serializer.is_valid(raise_exception=True)
-    return serializer.validated_data
+    return serializer.data
```
schenedx, line 225: Why change this to `.data`? What is wrong with `.validated_data`?

registrar/apps/enrollments/tasks.py
```diff
@@ -67,7 +65,13 @@ def list_program_enrollments(self, job_id, user_id, file_format, program_key):
 
 @shared_task(base=UserTask, bind=True)
 def list_course_run_enrollments(
-        self, job_id, user_id, file_format, program_key, course_key   # pylint: disable=unused-argument
+        self,
+        job_id,
+        user_id,
+        file_format,
+        program_key,
+        course_key,
```
schenedx: Can we please rename this to `internal_course_key` to differentiate more explicitly?

registrar/apps/enrollments/tests/test_tasks.py
```diff
@@ -172,6 +172,14 @@ class ListCourseRunEnrollmentTaskTests(ListEnrollmentTaskTestMixin, TestCase):
         COURSE_ENROLLMENT_INACTIVE,
     )
     mocked_get_enrollments_method = 'get_course_run_enrollments'
+    course_key = 'course-1'
```
schenedx: Again, please rename this to `internal_course_key`

schenedx (changes requested):
Just some minor variable renames. Rest LGTM. Thank you

registrar/apps/enrollments/data.py
jansenk: `.validated_data` only includes the data that you passed in and that it validated. It doesn't include any additional fields other than what you give it `.data` includes all of the data for the object, including the new `course_key` field

registrar/apps/enrollments/data.py
jansenk: i think renaming `course_key` `internal_course_key` here will help to clear it up. We need that to make the lms request. The caller of this function passes in the internal key, and then optionallly, some other key that it's using to identify the course in the csv.

jansenk: It may be an unnecessary decoupling, i could probably just call the `effective_course_key` `internal_course_key` and it would behave exactly the same, but I was thinking that at this level we shouldn't have to care about external keys, we just know the key to make the request with and a key to put on the csv

schenedx: This is where the external logic meets internal logic. It's good to make the variable name to be explicit so the next developer understand the key translation happens here.

registrar/apps/enrollments/serializers.py
```diff
@@ -43,6 +43,11 @@ class CourseEnrollmentSerializer(serializers.Serializer):
     status = serializers.ChoiceField(choices=COURSE_ENROLLMENT_STATUSES)
     account_exists = serializers.BooleanField()
 
+    def to_representation(self, instance):
```
kdmccormick: Why not just add `course_key` as a field on the serializer?

registrar/apps/enrollments/data.py
```diff
@@ -216,11 +217,12 @@ def get_course_run_enrollments(program_uuid, course_key, client=None):
         - HTTPError if there is an issue communicating with LMS
         - ValidationError if enrollment data from LMS is invalid
     """
-    url = _lms_course_run_enrollment_url(program_uuid, course_key)
+    url = _lms_course_run_enrollment_url(program_uuid, internal_course_key)
     enrollments = _get_all_paginated_results(url, client)
-    serializer = CourseEnrollmentSerializer(data=enrollments, many=True)
+    context = {'course_key': displayed_course_key or internal_course_key}
```
kdmccormick: The logic seems redundant here. 
* from caller: `displayed_course_key = external_course_key or internal_course_key`
* here: `course_key = displayed_course_key or internal_course_key`

Why not just take `external_course_key` as an optional parameter?

kdmccormick (changes requested):

registrar/apps/api/v1/mixins.py
```diff
@@ -154,14 +154,14 @@ class CourseSpecificViewMixin(ProgramSpecificViewMixin):
...
+    this mixin provides the `internal_course_key` and `external_course_key` properties,
+    which return the edX course key of the course referred to in the URL (either by the
     edX course key itself or by an external course key). It also validates
-    that they course key exists within the program, raising 404 if not.
+    that the course key exists within the program, raising 404 if not.
     """
 
     @cached_property
-    def internal_course_key(self):
+    def course_run(self):
```
kdmccormick, line 164: Nice refactor :+1:

registrar/apps/enrollments/data.py
kdmccormick: @jansenk Now that the serializer has `course_key` as a field, are you able to change this back to `validated_data`?

kdmccormick (approved):
I haven't manually tested it, but other than the one response I added to Simon's question, the code looks great :+1:

iloveagent57 (approved):
This looks good to me.  As others pointed out, nice refactoring.

registrar/apps/enrollments/data.py
jansenk: Doesn't look like it, still needs to be `.data`

PR #145, opened 2019-06-27
Aed/csrf

iloveagent57:
Couple of things we need to provide CSRF tokens from the registrar service to any frontends that need it (like the CSV manager).

### DIFF
--- a/registrar/settings/base.py
+++ b/registrar/settings/base.py
@@ -29,7 +29,9 @@ INSTALLED_APPS = (
 )
 
 THIRD_PARTY_APPS = (
+    'analytics',
     'corsheaders',
+    'csrf.apps.CsrfAppConfig',  # Enables frontend apps to retrieve CSRF tokens
     'guardian',
     'release_util',
     'rest_framework',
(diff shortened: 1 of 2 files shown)


registrar/settings/base.py
```diff
@@ -236,6 +238,9 @@
     'JWT_VERIFY_EXPIRATION': True,
     'JWT_VERIFY_AUDIENCE': False,
     'JWT_DECODE_HANDLER': 'edx_rest_framework_extensions.auth.jwt.decoder.jwt_decode_handler',
+    'JWT_AUTH_COOKIE': 'edx-jwt-cookie',
```
schenedx, line 241: I thought these values are all specified in various different environment settings file. 
Would these value conflict with those values in Stage and Prod? Does stage and prod settings just override these?

schenedx (approved):
Besides the JWT settings question, rest LGTM
I think we just need to test this on stage correctly.

registrar/settings/base.py
iloveagent57: The production settings file will update the `JWT_AUTH` dict from our prod/stage configuration, see this chunk of code: https://github.com/edx/registrar/blob/master/registrar/settings/production.py#L20-L35

schenedx: I thought it should be that way. Just wanted to confirm

PR #147, opened 2019-06-28
Add a trivial v2 API.

iloveagent57:
In support of releasing a new partner integration guide, we're releasing a v2 of the API that simply imports all of the URL patterns from v1.

https://openedx.atlassian.net/browse/EDUCATOR-4441

### DIFF
--- a/registrar/apps/api/v2/urls.py
+++ b/registrar/apps/api/v2/urls.py
@@ -1,0 +1,8 @@
+""" URL patterns for the v2 REST API. """
+
+from registrar.apps.api.v1.urls import urlpatterns as v1_urlpatterns
+
+
+app_name = 'v2'
+
+urlpatterns = [pattern for pattern in v1_urlpatterns]
--- a/.api-generated.yaml
+++ b/.api-generated.yaml
@@ -1325,6 +1325,400 @@ paths:
       summary: Enroll students in a program
       tags:
       - v1
+  /v2/jobs/{job_id}/:
+    get:
+      operationId: registrar_v2_get_job_status
+      parameters:
+      - description: UUID-4 job identifier
+        example: 3869c0d5-e88f-4088-bf1d-409222492869
+        format: uuid
+        in: path
+        name: job_id
+        required: true
+        type: string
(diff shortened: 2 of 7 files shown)


zacharis278 (approved):
💃

kdmccormick (changes requested):
Nice. Can you update `api.yaml` too?

kdmccormick:
@iloveagent57 also, can you update [this](https://github.com/edx/registrar/blob/master/registrar/apps/api/v1/mixins.py#L224) so that `job_url`s point to the V2 API?

schenedx (commented):
Do you need a change in api.yaml file as well?

registrar/apps/api/v2/urls.py
```diff
@@ -0,0 +1,8 @@
+""" URL patterns for the v2 REST API. """
+
+from registrar.apps.api.v1.urls import urlpatterns as v1_urlpatterns
+
+
+app_name = 'v2'
+
+urlpatterns = [pattern for pattern in v1_urlpatterns]
```
jansenk, line 8: nit: do you even need to do this? could you just do `urlpatterns = v1_urlpatterns`?

jansenk:
would we also want to make a change to the values of `event_method_map`? They specify v1

registrar/apps/api/v2/urls.py
iloveagent57: No, it's definitely not required - I left that as a sort of breadcrumb for when we have to pick and choose which patterns we want from v1 and which ones we need to explicitly define in v2.

iloveagent57:
Ooh, that `event_method_map` thing is a good point, so is that JobStatusMixin thing - I think this might be a little more involved than I initially thought.  I'm going to work on refactoring this out a little bit to see if I can make it more consistent regardless of which set of endpoints you actually hit.

schenedx (approved):
👍 LGTM. We would need some comprehensive testing for this on Stage, I'd say.

kdmccormick (approved):
Nice!

PR #150, opened 2019-07-03
Fix the bug where we should handle event_name do not exist

schenedx:
## Description
[EDUCATOR-4474](https://openedx.atlassian.net/browse/EDUCATOR-4474)
This is the change to fix the bug where if the HTTP method is not supported, we throw 500 error.
I also added a simple test case to make sure we test against this case where we should send 405 in this case.

@edx/masters-neem Please help review

### DIFF
--- a/registrar/apps/api/v1/tests/test_views.py
+++ b/registrar/apps/api/v1/tests/test_views.py
@@ -195,6 +195,37 @@ class S3MockMixin(object):
 
 
 @ddt.ddt
+class ViewMethodNotSupportedTests(RegistrarAPITestCase, AuthRequestMixin):
+    """ Tests for the case if user requested a not supported HTTP method """
+
+    method = 'DELETE'
+    path = 'programs'
+
+    @ddt.data(
+        ('programs', 'ProgramListView'),
+        ('programs/masters-in-english', 'ProgramRetrieveView'),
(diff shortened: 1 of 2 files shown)


jansenk (approved):
lgtm

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -194,6 +194,37 @@ def tearDownClass(cls):
         super().tearDownClass()
 
 
+@ddt.ddt
+class ViewMethodNotSupportedTests(RegistrarAPITestCase, AuthRequestMixin):
+    """ Tests for the case if user requested a not supported HTTP method """
+
+    method = 'DELETE'
```
kdmccormick, line 201: Nit: Changing this constant won't change the method used. I recommend either hard-coding it in, or using `self.request(self.method, ...`

kdmccormick (approved):
Nice fix! One non-critical nit.

registrar/apps/api/v1/tests/test_views.py
schenedx: The reason we need this constant is because the `AuthRequestMixin` have the unit test `test_unauthorized` that need method specified. 
I think using the `self.request(self.method...` is a decent approach

PR #153, opened 2019-07-08
allow csvs with extraneous fields

jansenk:
don't throw an error when there are extra fields, just make sure the required ones are there

### DIFF
--- a/registrar/apps/api/v1/tests/test_views.py
+++ b/registrar/apps/api/v1/tests/test_views.py
@@ -1774,10 +1774,10 @@ class EnrollmentUploadMixin(object):
         self.assertEqual(upload_response.status_code, 409)
 
     def test_enrollment_upload_invalid_header(self):
-        enrollments = [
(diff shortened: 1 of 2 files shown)


zacharis278 (approved):

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -1827,6 +1827,18 @@ def test_no_file(self):
             )
         self.assertEqual(upload_response.status_code, 400)
 
+    def test_extra_columns(self):
+        enrollment = self.build_enrollment('enrolled', '001')
+        enrollment['blood_type'] = 'AB-'
```
schenedx, line 1832: Interesting choice on the random extra column name

schenedx (approved):
👍

registrar/apps/api/v1/tests/test_views.py
jansenk: you'll like the ones in the other PR then too ;)

PR #155, opened 2019-07-11
Update CORs settings in registrar so Program Manager can hit it on stage

reillz10:
## Description


douglashall (approved):
You'll definitely need this. You may also need other things. I would take a look at how other backend service's CORS-related django settings are set. https://github.com/edx/ecommerce may be a good example.

PR #156, opened 2019-07-11
Update CORs settings in registrar so Program Manager can hit it on stage

reillz10:
## Description

### DIFF
--- a/registrar/settings/base.py
+++ b/registrar/settings/base.py
@@ -1,6 +1,8 @@
 import os
 from os.path import join, abspath, dirname
 
+from corsheaders.defaults import default_headers as corsheaders_default_headers
+
 from registrar.settings.utils import get_logger_config
 
 # PATH vars
@@ -68,6 +70,10 @@ MIDDLEWARE = (
 
 # Enable CORS
 CORS_ALLOW_CREDENTIALS = True
+CORS_ALLOW_HEADERS = corsheaders_default_headers + (
+    'use-jwt-cookie',
+)
+
 
 
 ROOT_URLCONF = 'registrar.urls'


registrar/settings/production.py
```diff
@@ -61,3 +63,7 @@
 if vars().get(jwk_setting) == 'None':
     vars()[jwk_setting] = None
 del jwk_setting
+
+CORS_ALLOW_HEADERS = corsheaders_default_headers + (
```
iloveagent57: Let's move this into the base settings file.  Both discovery and edx-platform define it there.

registrar/settings/base.py
```diff
@@ -67,7 +67,7 @@
 )
 
 # Enable CORS
-CORS_ALLOW_CREDENTIALS = True
+CORS_ALLOW_CREDENTIALS = False
```
iloveagent57: Other services (discovery and edx-platform) have this set to True.  Not sure where is better to copypasta from...

iloveagent57: Also, the comment above becomes a lie when we set this to False.

iloveagent57: https://github.com/ottoyiu/django-cors-headers/blob/a259cb18247f3ab4655a166483438772b3d66287/README.rst#cors_allow_credentials . I think we need to allow cookies in the cross-site requests for any JWT auth to work, right?

iloveagent57 (commented):
I'd set that back to true and move the headers thing into base.py.  Maybe we wanna double-check with @douglashall though.

douglashall:
> I'd set that back to true and move the headers thing into base.py. Maybe we wanna double-check with @douglashall though.

Sounds good to me!

registrar/settings/base.py
reillz10: Ha! the comment one is a funny miss on my end. I'm going to do all the things you said and Doug agreed with

registrar/settings/production.py
reillz10: Done

iloveagent57 (approved):

PR #157, opened 2019-07-12
command to create or modify programs

jansenk:
`./manage.py manage_programs <discovery_uuid>:<program_key>*`


kdmccormick:
Preliminary comment -- we wrote the Jenkins and Ansible not to pass the `org_key` into this command. If there's a way to avoid taking in the `org_key`, then that would save us from changing that work.

jansenk:
No longer require org_key, now we look at the authoring_organizations on the program and use the first one we also have in Registrar

zacharis278 (changes requested):
Looks good just one comment.  If we're gonna assign the first matching organization to every program can we add a comment to make that decision more obvious?

zacharis278 (approved):

PR #158, opened 2019-07-12
add trailing slash automatically

jansenk:
set append-slash to true


kdmccormick (changes requested):
Nice. Can you remove all the instances of `/?` that we added to `urls.py` files?

registrar/urls.py
```diff
@@ -34,8 +35,10 @@
 
 urlpatterns = oauth2_urlpatterns + [
     url(r'^admin/', admin.site.urls),
+    url(r'^admin$', RedirectView.as_view(pattern_name='admin:index')),
     url(r'^api/', include(api_urls)),
     url(r'^api-docs/', api_renderer.render_yaml_spec, name='api-docs'),
+    url(r'^api-docs', RedirectView.as_view(pattern_name='api-docs')),
```
kdmccormick: Why no `$` here?

kdmccormick (approved):
:cool:

PR #159, opened 2019-07-15
return 404 if discovery returns 404

jansenk:
If we 404 trying to get a program from discovery, return a 404 rather than a 500

### DIFF
--- a/registrar/apps/enrollments/data.py
+++ b/registrar/apps/enrollments/data.py
@@ -79,6 +79,7 @@ class DiscoveryProgram(object):
         Get a DiscoveryProgram instance, either by loading it from the cache,
         or query the Course Discovery service if it is not in the cache.
 
+        Raises Http404 if program is not cached and Discovery returns 404
         Raises HTTPError if program is not cached and Discover returns error.
         Raises ValidationError if program is not cached and Discovery returns
             data in a format we don't like.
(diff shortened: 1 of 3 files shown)


registrar/apps/enrollments/data.py
```diff
@@ -84,7 +86,13 @@ def get(cls, program_uuid, client=None):
         key = PROGRAM_CACHE_KEY_TPL.format(uuid=program_uuid)
         program = cache.get(key)
         if not (program and program.version == cls.class_version):
-            program = cls.load_from_discovery(program_uuid, client)
+            try:
```
jansenk: I should update the method docstring to reflect this

registrar/apps/enrollments/data.py
```diff
@@ -84,7 +86,13 @@ def get(cls, program_uuid, client=None):
...
+                program = cls.load_from_discovery(program_uuid, client)
```
schenedx: Is there a way you can put the `try...catch` block inside the function `load_from_discovery`?

schenedx (commented):
See my question inline

registrar/apps/enrollments/data.py
jansenk: actually, looking at this now, it would probably make most sense to put this all the way down in _make_request()

jansenk: actually that may be a heavier change than I thought ... i'll keep looking though

schenedx (approved):
👍

PR #161, opened 2019-07-17
Change course_key to course_id

kdmccormick:
Replace field names and code references to the field `course_key` with the field `course_id`


jansenk:
I looked over what you did already and only saw a single line that needed to change, so i just committed that

schenedx (approved):
👍 
Can you verify this change on stage once you merge?

PR #162, opened 2019-07-18
make argument single comma separated string

jansenk:
change signature from 
`./manage.py manage_programs uuid:key uuid:key uuid:key`
to 
`./manage.py manage_programs uuid:key,uuid:key,uuid:key`


kdmccormick (changes requested):
We also want to be able to handle programs without specified external keys. For example, we may want to pass in:
`uuid1:key1,uuid2,uuid3,uuid4:key4`
where uuids 2 and 3 have no external key.

In such cases, default the external key to `discovery_program.marketing_slug`

kdmccormick (approved):
:cool:

PR #163, opened 2019-07-22
Move newly added variables and edx_django_service variables to app repo

syedimranhassan:
## Description

### DIFF
--- a/registrar/settings/base.py
+++ b/registrar/settings/base.py
@@ -88,11 +88,12 @@ WSGI_APPLICATION = 'registrar.wsgi.application'
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.',
-        'NAME': '',
-        'USER': '',
-        'PASSWORD': '',
-        'HOST': '',  # Empty for localhost through domain sockets or '127.0.0.1' for localhost through TCP.
+        'NAME': 'registrar',
+        'USER': 'registrar001',
+        'PASSWORD': 'password',
+        'HOST': 'localhost',  # Empty for localhost through domain sockets or '127.0.0.1' for localhost through TCP.
         'PORT': '',  # Set to empty string for default.
+        'ATOMIC_REQUESTS': False,
     }
 }
 
@@ -146,7 +147,7 @@ BROKER_URL = "{0}://{1}:{2}@{3}/{4}".format(
 # Internationalization
 # https://docs.djangoproject.com/en/dev/topics/i18n/
 
-LANGUAGE_CODE = 'en-us'
+LANGUAGE_CODE = 'en'
 
 TIME_ZONE = 'UTC'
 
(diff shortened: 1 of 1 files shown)


registrar/settings/base.py
```diff
@@ -263,8 +273,8 @@
 # END AUTHENTICATION CONFIGURATION
 
 # Other service locations
-LMS_BASE_URL = 'replace-me'
```
kdmccormick: Why change these to `None`?

registrar/settings/base.py
```diff
@@ -278,7 +288,7 @@
 SEGMENT_KEY = None
 
 # Publicly-exposed base URLs for service and API
-API_ROOT = 'http://replace-me/api'
```
kdmccormick, line 281: As everything else in here is based on `127.0.0.1:8000`, can you make this `http://127.0.0.1:8000/api`?

registrar/settings/base.py
```diff
@@ -290,3 +300,25 @@
...
+    'DEFAULT_FILE_STORAGE': 'django.core.files.storage.FileSystemStorage',
+    'MEDIA_ROOT': MEDIA_ROOT,
+    'MEDIA_URL': MEDIA_URL
+}
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
+    }
+}
+SOCIAL_AUTH_EDX_OIDC_KEY = 'registrar-key'
```
kdmccormick: As I understand it, edX OpenIDConnect is deprecated, so would it be safe for us to nix the `SOCIAL_AUTH_EDX_OIDC_` settings entirely?

registrar/settings/base.py
```diff
@@ -290,3 +300,25 @@
 REGISTRAR_SERVICE_USER = 'registrar_service_user'
 SESSION_EXPIRE_AT_BROWSER_CLOSE = False
 STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.StaticFilesStorage'
+CORS_ORIGIN_WHITELIST = []
+CSRF_TRUSTED_ORIGINS = []
+MEDIA_STORAGE_BACKEND = {
```
kdmccormick: `MEDIA_STORAGE_BACKEND` is not a real Django setting; it's just something we use in our configuration pipeline, so you can remove this dictionary.

registrar/settings/base.py
```diff
@@ -290,3 +300,25 @@
 REGISTRAR_SERVICE_USER = 'registrar_service_user'
 SESSION_EXPIRE_AT_BROWSER_CLOSE = False
 STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.StaticFilesStorage'
+CORS_ORIGIN_WHITELIST = []
```
kdmccormick: Can you put `CORS_ORIGIN_WHITELIST` with the other CORS settings (line ~76)?

kdmccormick (changes requested):
Generally looks good, just a few comments.

registrar/settings/base.py
coryleeio: We're unsure - if you say it is we can remove it.  We are just merging the default values from configuration into this repo in preparation to delete them from the configuration repo.

If you say it is okay to remove it we can remove it, let us know if you want it removed.

coryleeio (approved):
This is good on my end - but we should address Kyle's feedback before merging.

registrar/settings/base.py
kdmccormick: Sounds good. I did some digging... you're good to drop all the `SOCIAL_AUTH_EDX_OIDC_*` settings @syedimranhassan

PR #169, opened 2019-08-20
Ignore read-only/old jobs when checking conflicts

kdmccormick:
@edx/masters-neem 

**Motivation**
Curtin will be not be able to run new enrollment-writing jobs due to two enrollment-reading jobs that are currently stuck in the Pending state.

**Description**
Currently, we disallow any enrollment-writing job from starting if the associated program has any
jobs in the one of processing states (IN PROGRESS, RETRYING, and PENDING). This commit relaxes that constraint in two ways:
* Only block if a 'write' job is processing
* Ignore jobs of a certain age (currently 30 minutes)

The reason for the second relaxation is that jobs ocassionally becomes 'stuck' in the PENDING state, so
we have to modify the database to unblock future jobs for that program. This is not desirable, so we can assume that any job older than 30 minutes will never complete.

If Registrar begins to have longer-running jobs, then the maximum age may need to be reconsidered.

### DIFF
--- a/registrar/apps/api/v1/tests/test_views.py
+++ b/registrar/apps/api/v1/tests/test_views.py
@@ -1786,7 +1786,10 @@ class EnrollmentUploadMixin(object):
     def test_enrollment_upload_conflict(self):
         enrollments = [self.build_enrollment('enrolled', '001')]
 
-        with mock.patch('registrar.apps.api.v1.views.is_enrollment_job_processing', return_value=True):
+        with mock.patch(
+                'registrar.apps.api.v1.views.is_enrollment_write_blocked',
+                return_value=True
+        ):
             upload_response = self._upload_enrollments(enrollments)
 
         self.assertEqual(upload_response.status_code, 409)
--- a/registrar/apps/api/v1/views.py
+++ b/registrar/apps/api/v1/views.py
@@ -54,7 +54,7 @@ from registrar.apps.enrollments.tasks import (
     write_course_run_enrollments,
     write_program_enrollments,
 )
-from registrar.apps.enrollments.utils import is_enrollment_job_processing
+from registrar.apps.enrollments.utils import is_enrollment_write_blocked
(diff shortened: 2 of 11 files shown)


zacharis278:
wondering if it would make more sense to just set a timeout on our celery tasks instead of ignoring jobs that never complete?

kdmccormick:
@zacharis278 that does seem like a nicer solution, assuming the timeout could mark task as Failed.

zacharis278:
Yeah unfortunately my solution is unlikely to fix the current state.  If we're unsure about the cause and if the timeout response will get persisted.  I'd recommend putting in both solutions so we can track the behavior without blocking partners while we fix the DB.  If the timeout does work, we can can clean it up to rely on that going forward.  Rather not have a bunch of mystery pending jobs around that may or may not be holding up workers.

kdmccormick:
@Zacharis278 Fortunately, the only stuck jobs currently are Read jobs:
```
mysql> select name, state from user_tasks_usertaskstatus where state in ('Pending', 'Retrying', 'In Progress');
+-------------------------------------------------------------------------------+---------+
| name                                                                          | state   |
+-------------------------------------------------------------------------------+---------+
| online-master-in-marketing-curtin-university:read:list_program_enrollments    | Pending |
| online-master-in-marketing-curtin-university:read:list_course_run_enrollments | Pending |
+-------------------------------------------------------------------------------+---------+
2 rows in set (0.00 sec)
```
So, I could implement just the first part of this (ignore read-only jobs) to fix the existing issue, and then use Celery timeouts to prevent more jobs from getting stuck.

zacharis278 (approved):
@kdmccormick sounds good.  Don't think there is a good reason to block on readonly anyway.

kdmccormick:
Closed in favor of https://github.com/edx/registrar/pull/170 + https://github.com/edx/registrar/pull/171

PR #171, opened 2019-08-20
Kill Celery tasks after five minutes

kdmccormick:
@edx/masters-neem 

Set `CELERY_TASK_SOFT_TIME_LIMIT` to four minutes and `CELERY_TASK_TIME_LIMIT` to five minutes. This should prevent enrollment-writing UserTasks from getting stuck indefinitely in
a processing state and blocking future tasks.

I know no way of testing this, automated or manually. Unless there are any suggestions, we'll just have to keep an eye on "Job already in progress for program" errors.


zacharis278 (approved):
🤞

PR #173, opened 2019-08-22
Handle whitespace and Windows newlines in CSVs

kdmccormick:
@edx/masters-neem 
https://openedx.atlassian.net/browse/EDUCATOR-4580
https://openedx.atlassian.net/browse/EDUCATOR-4585

[EDUCATOR-4575](https://openedx.atlassian.net/browse/EDUCATOR-4575) was caused in part by the fact that our CSV processing didn't tolerate extra hitespace around CSV values.

This commit desensitizes our CSV processing in several ways:
* Leading, trailing, and padding whitespace will
  be ignored in all rows, including the header.
* Empty rows will be ignored.
* Headers will always be lowercased.
* Windows-style (\r\n) and Unix-style (\n) line-endings
  will both be tolerated, even within the same file.

Finally, extra columns (ones not specified in field_names) will be dropped from loaded CSVs. This is to avoid deserializion issues in the future.


jansenk (approved):
👍

iloveagent57 (approved):

PR #175, opened 2019-08-28
Update BROKER_URL setting from config in production.py (OPS-4272)

nadeemshahzad:
## Description

We are removing extra variables from environment of services and adding to config file (https://github.com/edx/configuration/pull/5381). In registrar celery settings are going to config and  ```BROKER_URL``` need to be build from variables injected from config in production.py rather than importing this setting from base.py.

### DIFF
--- a/registrar/settings/production.py
+++ b/registrar/settings/production.py
@@ -46,6 +46,10 @@ DB_OVERRIDES = dict(
 for override, value in DB_OVERRIDES.items():
     DATABASES['default'][override] = value
 
-CELERY_ALWAYS_EAGER = (
-    os.environ.get("CELERY_ALWAYS_EAGER", "false").lower() == "true"
+BROKER_URL = "{0}://{1}:{2}@{3}/{4}".format(
+    CELERY_BROKER_TRANSPORT,
+    CELERY_BROKER_USER,
+    CELERY_BROKER_PASSWORD,
+    CELERY_BROKER_HOSTNAME,
+    CELERY_BROKER_VHOST
 )


registrar/settings/production.py
```diff
@@ -49,3 +49,11 @@
 CELERY_ALWAYS_EAGER = (
     os.environ.get("CELERY_ALWAYS_EAGER", "false").lower() == "true"
 )
+
```
kdmccormick: Please remove the `CELERY_ALWAYS_EAGER = ( os.environ.get ...` statement above, because that value will also be injected by config now once https://github.com/edx/configuration/pull/5381 merges.

kdmccormick (commented):
Cool, thanks for simplifying this. Just one request, but looks good otherwise.

As a side note, this creates some inconsistency between how we load Celery settings in `base.py` vs `production.py`, but that's something the Master's team can take care of in one sweep once the config simplification is done.

registrar/settings/production.py
nadeemshahzad: Thanks @kdmccormick , removed ```CELERY_ALWAYS_EAGER``` to get this value injected from config.

kdmccormick (approved):
LGTM, but I'm going to ask someone else on my team to also look at this.

kdmccormick:
FYI for onlookers: created https://openedx.atlassian.net/browse/EDUCATOR-4605 to unify how we handle these settings (and other similar ones) between `base.py` and `production.py`

iloveagent57 (approved):
LGTM

PR #187, opened 2019-09-24
Fix the issue where if a user has global_permission, no access to programs

schenedx:
We are seeing an issue that if a user have global permissions, the user cannot see programs on program manager

### DIFF
--- a/registrar/apps/api/v1/views.py
+++ b/registrar/apps/api/v1/views.py
@@ -89,10 +89,21 @@ class ProgramListView(AuthMixin, TrackViewMixin, ListAPIView):
             programs = programs.filter(
                 managing_organization=self.organization_filter
             )
-        if self.permission_filter:
+        if not self.permission_filter:
+            return programs
+
+        # if the user has permissions across organizations
+        # via membership in a "global-access" group, give them
+        # access to all programs
+        user = self.request.user
+        if user.has_perm(self.permission_filter):
+            return programs
+        else:
+            # otherwise, check if the user has the required permissions
+            # within the organization for each program
             programs = (
                 program for program in programs
-                if self.request.user.has_perm(
+                if user.has_perm(
                     self.permission_filter, program.managing_organization
                 )
             )
--- a/registrar/apps/api/v1/tests/test_views.py
+++ b/registrar/apps/api/v1/tests/test_views.py
@@ -40,6 +40,7 @@ from registrar.apps.core.jobs import (
 from registrar.apps.core.models import Organization, OrganizationGroup
 from registrar.apps.core.permissions import JOB_GLOBAL_READ
 from registrar.apps.core.tests.factories import (
+    GroupFactory,
     OrganizationFactory,
     OrganizationGroupFactory,
     UserFactory,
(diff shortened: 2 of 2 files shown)


registrar/apps/api/v1/views.py
```diff
@@ -89,10 +89,11 @@ def get_queryset(self):
             programs = programs.filter(
                 managing_organization=self.organization_filter
             )
-        if self.permission_filter:
+        user = self.request.user
+        if self.permission_filter and not user.has_perm(self.permission_filter):
```
kdmccormick: Nice. Consider adding a comment here explaining that we check global permissions first, and then program-specific permissions.

schenedx: Will do. Also, can you review my unit test addition?

iloveagent57: I would structure it a little differently to make it really clear what's going on:
```python
user = self.request.user
# if the user has permissions across organizations
# via membership in a "global-access" group, give them
# access to all programs
if user.has_perm(self.permission_filter):
    return programs
if self.permission_filter:
    # otherwise, check if the user has the required permissions
    # within the organization for each program
    programs = (
        program for program in programs
        if user.has_perm(
            self.permission_filter, program.managing_organization
        )
    )
return programs
```

iloveagent57: The reason I like structuring it as in my comment is that it makes it _really_ clear that a user having global permissions is a special case.

iloveagent57 (changes requested):
This is obviously fine from a functional perspective, I have a recommendation about code clarity.

iloveagent57 (approved):
please squash and merge

registrar/apps/api/v1/views.py
```diff
@@ -89,10 +89,18 @@ def get_queryset(self):
             programs = programs.filter(
                 managing_organization=self.organization_filter
             )
+        user = self.request.user
+        # if the user has permissions across organizations
+        # via membership in a "global-access" group, give them
+        # access to all programs
+        if user.has_perm(self.permission_filter):
+            return programs
```
kdmccormick, line 100: `self.permission_filter` will be `None` if the `user_has_perm` query param is not provided. If this is working, it's probably because `user.has_perm(None)` is incidentally evaluating to `False`.

If you are going to check global permissions in its own branch, do
```python
if self.permission_filter and user.has_perm(self.permission_filter):
    ...
```

kdmccormick (changes requested):

registrar/apps/api/v1/views.py
schenedx: @kdmccormick I specifically check this point. The function `user.has_perm(None)` is handled with always `false` returned. I checked this using pdb. Isn't this what we want anyways?

schenedx: Updated. Please review again

kdmccormick: from Slack conversation: return value of `user.has_perm(None)` is not documented, it returns `True` for me, so it's best to check explicitly.

iloveagent57: how about an early return?
```python
if not self.permission_filter:
    return programs
if user.has_perm(self.permission_filter):
   # ...etc.
```

kdmccormick: Would probably be the most elegant, but I know @schenedx is trying to get this out the door ASAP

kdmccormick (approved):

PR #189, opened 2019-10-18
Deprecate & Remove Registrar v1-mock

how015:
@edx/masters-devs 

ticket# : EDUCATOR-4676

Description:
Registrar's v1-mock is never used and should be removed.


kdmccormick:
The code changes are looking good so far!

I'm not sure why the checks are failing... I don't think it has to do with your changes. I'll investigate.

In the mean time, I remembered that we need to update our Amazon API proxy config as well. It should be as simple as removing these lines from the api-manager repository: https://github.com/edx/api-manager/blob/master/swagger/api.yaml#L109-L119

kdmccormick:
@how015 Also, you'll want to re-generate `.api-generated.yaml` and commit the changes. [This part of the README](https://github.com/edx/registrar#api-documentation) explains more.

kdmccormick (commented):
Good job. However, the coverage changes are related to your PR.

The "codecov/patch" passed because _every line of code you modified is covered by tests._

However, by removing v1-mock, there are now lines that _are no longer run_. Even though you didn't change them, you changed code that referenced them, and they are no longer covered. That's why "codevcov/project" is failing.

If you click into "codecov/project" **Details**, you can see exactly which lines are no longer covered.

To keep the check at 100%, you can:
* Delete the `name_to_key` function and its related imports
* Mark the `post` function with `pragma: no cover`, as we're not using it, but it's a good function to keep around.

PR #192, opened 2019-10-18
[Fix upgrades attempt 1] Clean up requirements constraints; include .txt files

kdmccormick:
@edx/masters-devs 

# Context

After fixing travis builds, our `make upgrade` was broken.
https://openedx.atlassian.net/browse/EDUCATOR-4760

# This PR

The changes in this pr are inspired by [this blog post on a successful pip-tools workflow](https://jamescooke.info/a-successful-pip-tools-workflow-for-managing-python-package-requirements.html):
* Generate all `.txt` files, not just ones needed for installing.
* For all `.in` files, include `.txt` files instead of `.in` files. This makes the pins in all of our `.txt` files compatible with each other. So, for example, `local.txt` can not longer pin `package==1.0` if `production.txt` pins `package==1.1`.
* Moves all constraints to `constraints.txt` and adds reasons for each
* Add `-c constraints.txt` to every `.in` file
* Add convenient Makefile rules
* Runs `make upgrade` in a separate commit

# Problem

`make upgrade` runs, but the generated `.txt` files don't include the dependencies of `pylint`.
So, running `make validate` fails when trying to run pyint.


kdmccormick:
Closed in favor of #197

PR #194, opened 2019-10-21
[Fix upgrades attempt 2] Generate all requirements at once into pins.txt

kdmccormick:
@edx/masters-devs 

# Context

After fixing travis builds, our `make upgrade` was broken.
https://openedx.atlassian.net/browse/EDUCATOR-4760

# This PR
In my quest to fix `make upgrade`, I tried changing how we use pip-tools.

The pip-tools docs recently started giving recommendations on how to do layered requirements files:
https://github.com/jazzband/pip-tools#workflow-for-layered-requirements

This does a modified version of that:
* constraints.in holds all manual version constraints.
* pins.txt holds all exact version pins.
* .env.in files specify lists of packages needed for various environments (devstack, production, etc.).
  They are constrained by pins.txt and used to compile .env.txt files, all of which are subsets of pins.txt.
* pins.in includes all the environments, is constrained  by constraints.in, and is used to generate pins.txt.


kdmccormick:
Closed in favor of #197

PR #197, opened 2019-10-25
[Fix upgrades attempt 5] Just pin edx-lint==1.3.0 in test.in

kdmccormick:
@edx/masters-devs 

After fixing travis builds, our `make upgrade` was broken (https://openedx.atlassian.net/browse/EDUCATOR-4760)

This PR simply pins `edx-lint==1.3.0` in `test.in`, which isn't the "right" place, but whatever.

Follow-up to un-pin edx-lint: https://openedx.atlassian.net/browse/EDUCATOR-4764


matthugs (approved):

schenedx (approved):
👍

PR #199, opened 2019-10-29
Expose the program type field to the API on Program endpoint

schenedx:
## Description

EDUCATOR-4667 This is the PR to expose the program type field out through the API
This is needed to help program manager to list out the correct UI

@edx/masters-devs Please review


zacharis278 (approved):
💯

PR #203, opened 2019-11-08
Added devstack config yml file to app repo

syedimranhassan:
## Description

### DIFF
--- a/registrar/settings/registrar.yml
+++ b/registrar/settings/registrar.yml
@@ -1,6 +1,4 @@
 ---
-# Ansible managed
-
 
 API_ROOT: http://localhost:18734/api
 BACKEND_SERVICE_EDX_OAUTH2_KEY: registrar-backend-service-key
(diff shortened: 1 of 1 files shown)


kdmccormick (commented):
Could you give some more context on the new file @syedimranhassan ? For example, where is the file loaded, and what is it a replacement for?

syedimranhassan:
@kdmccormick working on [this](https://openedx.atlassian.net/browse/OPS-4397) ticket. We are in process of removing rendering code tasks. So, we should hard code file in devstack so it'll not break.

registrar/settings/registrar.yml
```diff
@@ -0,0 +1,82 @@
+---
+# Ansible managed
```
kdmccormick: can you remove this comment?

kdmccormick (commented):
Thanks for the context. As far as I know, we don't actually use the rendered YAML for devstack -- we just declare all settings manually in https://github.com/edx/registrar/blob/master/registrar/settings/devstack.py . So I'm not sure that this file would actually be used anywhere.

However if you think this should still be merged because it's part of the larger config-simplification plan, go ahead.

registrar/settings/registrar.yml
syedimranhassan: @kdmccormick I have addressed your comment. Can you please review it?

kdmccormick (approved):
Sorry for the delay. I maintain that this file won't actually do anything currently, but I put a ticket in our backlog to change that: https://openedx.atlassian.net/browse/EDUCATOR-4802

fyi @schenedx

PR #206, opened 2019-11-20
Program Reports and AuthZ ADR

zacharis278:
@edx/masters-devs


docs/decisions/0004-program-authz.rst
```diff
@@ -0,0 +1,25 @@
+Program Level Authorization
+===========================
+
+Context
+-------
+
+As the scope of programs in the Registrar service expands we run into multiple cases where granting a user access to all programs within an organization is not desirable. An organization composed of two programs, "A" and "B", may wish to assign certain administrators access to only "A" or only "B". To solve this, we will expand our permission model to allow permissions to be assigned to a specific program.
```
schenedx: `administrators access to only "A" or only "B", but not the other`

how015 (approved):
LGTM

schenedx (commented):
This is largely what we talked about. That said, I still don't fully understand how the `programorggroup` object will be enforcing the reporting APIs.

docs/decisions/0004-program-authz.rst
```diff
@@ -0,0 +1,25 @@
...
+
+Decision
+--------
+1. We will expand on the existing role-based access control model outlined in 0003-authnz_ to add a new type of access control group called ``ProgramOrgGroup``.
+
+2. A ``ProgramOrgGroup`` will be associated with a single Program. Providing access to that ``Program`` resource and any resource that falls under it. (e.g. ``Courses`` and ``Enrollments``)
+
+3. A ``ProgramOrgGroup`` will also be associated with an ``Organization``. This does not provide access but acts as a reference to the ``Organization`` granting permissions to the requested ``Program``. This ``Organization`` value must be one the program's authoring organizations.
+
+4. When a request is made for a ``Program`` two permission objects will be checked for the required role: the requested ``Program`` and that program's `authoring_organization`. If either of these objects have the required role the request is considered authorized.
```
kdmccormick: When doing authorization checking, we don't look at roles; we look at permissions. An action can require multiple permissions of a user, and a role can provide multiple permissions to a user. Furthermore, based on this ADR, we're heading the direction of a program having more than one `authoring_organization`. 

For a given user, do we take the union the permissions assigned for each authoring org, and then union that with the permissions assigned for the program? Things are getting complicated, and I'm not sure how these pieces will all interact.

docs/decisions/0005-program-reports.rst
```diff
@@ -0,0 +1,32 @@
...
+
+As it stands today, Registrar is limited to the management of enrollments for master's programs. We wish to widen this context and act as an interface for programs of all types. Much of business logic required to setup and interact with a program already exists in this service. A second driving factor in this decision is the existing role-based access control (RBAC) in Registrar which allows us to grant users permissions at specific entities. This existing system can easily be modified to fit our needs for managing access to program reports (see 0004-program-authz_). By expanding the scope of Registrar we can keep program management code in one place as well as be the one stop shop for configuring role based access to program data.
```
matthugs: You don't abbreviate roll-based access control later on in this document, so letting folks know how you would abbreviate it is unnecessary

matthugs (commented):
Records! Visibility! 😻

docs/decisions/0004-program-authz.rst
```diff
@@ -0,0 +1,26 @@
...
+
+Decision
+--------
+1. We will expand on the existing role-based access control model outlined in 0003-authnz_ to add a new type of access control group called ``ProgramOrgGroup``.
+
+2. A ``ProgramOrgGroup`` will be associated with a single Program. Providing access to that ``Program`` resource and any resource that falls under it. (e.g. ``Courses`` and ``Enrollments``)
+
+3. A ``ProgramOrgGroup`` will also be associated with an ``Organization``. This does not provide access but acts as a reference to the ``Organization`` granting the role. This ``Organization`` value must be one the program's managing organizations.
+
+4. A user's permissions at a requested object will be defined as the union of all permissions granted on that object and all parents of that object.  In the case of a ``Program`` request, we must look at both the ``Program`` and its ``managing_organization`` to determine permissions.
```
kdmccormick: In the previous point you mention managing organizations (plural). How about `... and all of its managing organizations to determine ...`?

zacharis278:
Thanks for the feedback, I've made some updates based on that and our working group convo yesterday.

@schenedx I largely tried to keep the reporting feature out of this since `ProgramOrgGroup` could be used to grant any new or existing role.  In the case of reporting endpoints it's just a matter of creating a `ProgramOrgGroup` with a role that includes whatever permission the reporting endpoints require.

docs/decisions/0005-program-reports.rst
```diff
@@ -0,0 +1,32 @@
...
+--------
+1. Registrar, and by extension the Program Manager frontend, will act as the access mechanism for retrieving program report files.
+
+2. A new reporting app will be added to the service that will expose endpoints to list and retrieve available reports.
+
+3. We will add a ``type`` field to the ``Program`` model so MicroMasters and Professional Certificate programs may be configured in the service at the request of the partner.
+
+4. Endpoints for creating or retrieving enrollments will now require the requested program be of type `Masters` in addition to checking a user's permissions.
+
+5. A new role will be added to grant access to program reports. Existing roles will be refactored to be specific to either enrollment or reporting allowing us to configure access to those features separately.
```
kdmccormick, line 20: This may not affect the ADR, but something to note: Grading also uses the `READ_ENROLLMENTS` permission. The tickets that come out of this might be an opportunity to make a separate `READ_GRADES` permission.

kdmccormick (approved):
Approved, pending a couple comments.

docs/decisions/0004-program-authz.rst
zacharis278: fair point

docs/decisions/0005-program-reports.rst
zacharis278: humm maybe? If the only use case is one-to-one right now it may just be simpler to make that change if/when that is no longer true.

kdmccormick: makes sense. we can make that call when grooming the implementation tickets.

PR #207, opened 2019-11-21
Prevent org level enrollment permitted users to access enrollment api for Micromasters programs

schenedx:
## Description
Currently, we have quite a few users on registrar who can read and write the enrollments associated with programs within an organization. However, we should deny any user trying to access enrollment apis for those Micromasters programs.

### DIFF
--- a/registrar/apps/api/v1/mixins.py
+++ b/registrar/apps/api/v1/mixins.py
@@ -1,6 +1,7 @@
 """
 Mixins for the public V1 REST API.
 """
+import logging
 import uuid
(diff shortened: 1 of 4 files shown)


kdmccormick (commented):
It's hard to articulate exactly why, but despite my agreement in the meeting yesterday, this still feels hacky to me.

I'd like to talk more at a whiteboard today if we can. I still don't think I fully understand the issue.

schenedx:
closing this because of the better approach on https://github.com/edx/registrar/pull/209

PR #209, opened 2019-11-25
Do not allow enrollments API endpoints to be accessed for non-degree programs

schenedx:
## Description
Currently, we have quite a few users on registrar who can read and write the enrollments associated with programs within an organization. However, we should deny any user trying to access enrollment apis for those Micromasters programs.

### DIFF
--- a/registrar/apps/api/v1/mixins.py
+++ b/registrar/apps/api/v1/mixins.py
@@ -231,6 +231,7 @@ class EnrollmentMixin(ProgramSpecificViewMixin):
     """
     This mixin defines the required permissions
     for any views that read or write program/course enrollment data.
+    Overrides AuthMixin.check_permissions.
     """
     def get_required_permissions(self, request):
         if request.method == 'GET':
@@ -238,6 +239,17 @@ class EnrollmentMixin(ProgramSpecificViewMixin):
         if request.method == 'POST' or self.request.method == 'PATCH':
             return [perms.ORGANIZATION_WRITE_ENROLLMENTS]
         return []  # pragma: no cover
+
+    def check_permissions(self, request):
+        if not self.program.is_enrollment_enabled:
+            # Raise exception if the program (MM at the moment) is not
+            # available for enrollments related API endpoints
+            raise PermissionDenied(
+                'Cannot access enrollment endpoints with program [%s] whose enrollments are disabled',
+                self.program.key
+            )
+
+        super().check_permissions(request)
 
     def handle_enrollments(self, course_id=None):
         """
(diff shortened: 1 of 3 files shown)


registrar/apps/api/v1/mixins.py
```diff
@@ -239,6 +239,13 @@ def get_required_permissions(self, request):
             return [perms.ORGANIZATION_WRITE_ENROLLMENTS]
         return []  # pragma: no cover
 
+    def check_permissions(self, request):
+        super().self_permissions(self, request)
```
kdmccormick: `super().check_permissions`

registrar/apps/api/v1/mixins.py
```diff
@@ -239,6 +239,13 @@ def get_required_permissions(self, request):
             return [perms.ORGANIZATION_WRITE_ENROLLMENTS]
         return []  # pragma: no cover
 
+    def check_permissions(self, request):
```
kdmccormick, line 243: I'd mention in the docstring that you're overriding this from within DRF

kdmccormick (commented):
This approach looks great :+1:

registrar/apps/api/v1/mixins.py
```diff
@@ -239,6 +239,13 @@ def get_required_permissions(self, request):
...
+        super().self_permissions(self, request)
+        if not self.program.is_enrollment_enabled:
```
kdmccormick, line 244: I recommend flipping the order of these -- check if enrollment is enabled, then call super. Take care of the simple case first.

how015 (approved):
LGTM

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -999,7 +1011,7 @@ def test_discovery_404(self):
                 self.stem_admin,
                 req_data,
             )
-        self.assertEqual(404, response.status_code)
+        self.assertEqual(422, response.status_code)
```
kdmccormick, line 1014: Nice catch!

registrar/apps/api/v1/mixins.py
```diff
@@ -239,6 +240,14 @@ def get_required_permissions(self, request):
             return [perms.ORGANIZATION_WRITE_ENROLLMENTS]
         return []  # pragma: no cover
 
+    def check_permissions(self, request):
+        if not self.program.is_enrollment_enabled:
+            # Raise exception if the program (MM at the moment) is not
+            # available for enrollments related API endpoints
+            raise PermissionDenied()
```
kdmccormick: Consider putting a message in the exception; could help a future dev who's looking at Splunk or something.

kdmccormick (commented):
Nice.

However, I'm not sure this will work for CSV uploads, as those views don't inherit from `EnrollmentMixin`.

registrar/apps/api/v1/mixins.py
schenedx: Good point. I'll try to add proper logging here.

kdmccormick (approved):
LGTM pending exception message :+1:  (CSV uploads will be addressed in next PR)

PR #210, opened 2019-11-26
Move Program Model to Core [Part 1 of 3]

zacharis278:
@edx/masters-devs 

In order to assign new/existing guardian groups to Programs we need to move the model to the core app.  To reduce risk in this change I've split the work into 3 parts we can deploy and test in sequence.

### rollout plan
1. Move program model from enrollments to core but point to the existing database table. This migration should be a no-op and make no db changes. No other code/imports has been refactored yet.
2. Refactor imports and tests to remove references to Program as an enrollments mode.  Technically this could be combined with #1 but I chose to break it out since it generates a messy diff.
3. Update the table name referenced by the model to ``core_enrollments`` this will be the only database change.


zacharis278:
on second thought, let's just do this in 2 PRs it's not that complicated

PR #211, opened 2019-11-26
Move Program Model to Core [Part 1 of 2]

zacharis278:
@edx/masters-devs 

In order to assign new/existing guardian groups to Programs we need to move the model to the core app.  ~To reduce risk in this change I've split the work into 3 parts we can deploy and test in sequence.~  (changed my mind to just do 2 requests, this isn't really all that complicated)

### rollout plan
1. Move program model from enrollments to core and refactor imports but continue to use the existing database table. This migration should be a no-op and make no db changes.
3. Update the table name referenced by the model to ``core_enrollments`` this will be the only database change.


schenedx (commented):
I suspect you need another migration to add a new db table called core_programs with no data in it. Then you would need data migration done after this PR, right? Or are you just going to keep the database table name to be `enrollments_program` forever?

zacharis278:
@schenedx I'd still like to rename the table so it's correct in the end.  IMO that should be separate set of PR's and how we handle it doesn't block getting this model moved in code.

matthugs:
@Zacharis278 but step 2 would be blocked on getting the data moved/duplicated from the `enrollments_` table to the `core_` one, right? It seems like that'd be necessary to do directly before what you have written as step 2; could you clarify in the breakdown of the plan?

zacharis278:
Wanted to rename the table in place to avoid needing to move/duplicate the table data but this isn't going to work in a clean way.  Closing this in favor of a different approach

PR #212, opened 2019-11-26
Move Program Model to Core [Part 2 of 2]

zacharis278:
@edx/masters-devs 

Part 2 for migrating Program model to the core app.  See rollout plan here: https://github.com/edx/registrar/pull/210

_note: I've pointed the base brach to part 1 of this so the diffs is easier to read. I'll update it once we're ready to go_


registrar/apps/core/migrations/0007_rename_program_table.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0006_program_move'),
+    ]
+
+    operations = [
+        migrations.AlterModelTable(
```
schenedx, line 15: I think this migration will introduce problem. Because in your first PR, the model is separated from the DB tables. When you do this, the model will have no corresponding database table and you will introduce errors to the ORM. Have you tested this change on devstack?

zacharis278: shoot, you have a point.  This just works on devstack since deleting the ``db_table prop`` happened alongside the migration.  I guess the only way to rename this is to create a second copy of that table first.

PR #214, opened 2019-12-03
Create new program model in core app

zacharis278:
@edx/masters-devs

In order to assign new/existing guardian groups to Programs we need to move the model to the core app.

### rollout plan
1. Add new program model to core app (this PR) and notify admin users of the in-progress model change.
2. Manually copy over programs since there are only 9 read only records.
3. Second PR will be opened to point code at new model. (#215)
4. Cleanup old model.


registrar/apps/enrollments/migrations/0004_prepare_remove_organization.py
```diff
@@ -9,6 +9,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
+        ('core', '0002_organization_organizationgroup'),
```
schenedx, line 12: Why do we need this change?

schenedx (approved):
LGTM

registrar/apps/enrollments/migrations/0004_prepare_remove_organization.py
zacharis278: Just cleanup of a missing dependency. If you add new migrations that cause the order to change, this migration (``0004_prepare_remove_organization``) can break since it references a model created by ``0002_organization_organizationgroup``

matthugs (approved):
Expressed my concerns over slack about non-production environments & the difficulty of moving those carefully up to newer versions of registrar.

Thank you for the care you're putting into this transition!

registrar/apps/core/models.py
```diff
@@ -7,6 +7,7 @@
 from guardian.shortcuts import remove_perm
 from model_utils.models import TimeStampedModel
 
+from registrar.apps.common.data import DiscoveryProgram
```
kdmccormick, line 10: So now `common` depends on `core` and `core` depends on `common` :/

How much work would merging `common` and `core` be?

zacharis278: tbh I'm not sure why we have both.  Common has no models so merging would be pretty easy.

kdmccormick: https://github.com/edx/edx-platform/pull/22211

kdmccormick: https://openedx.atlassian.net/browse/MST-76

PR #216, opened 2019-12-03
Prevent MM programs showing on Program Manager

schenedx:
## Description
Currently, regardless of program types, the program managers would show the program as long as it is written into the programs model in registrar. We should filter out MM programs appropriately.

This PR would filter out MM programs if the requested perm_filter is of enrollment related. 

@edx/masters-devs Please help review.


zacharis278 (approved):

registrar/apps/api/v1/views.py
```diff
@@ -107,6 +108,8 @@ def get_queryset(self):
                     self.permission_filter, program.managing_organization
                 )
             )
+            if self.permission_filter in ENROLLMENT_PERMISSIONS_LIST:
+                programs = [program for program in programs if program.is_enrollment_enabled]
```
kdmccormick: I recommend doing this filter for global *and* org-level permissions, because even if I have global program-write permissions, I still can't write to MicroMaster's.

Suggested implementation:
* Remove `if user.has_perm...` and its early return
* Change the `else` to `if not user.has_perm...`
* Dedent your new filter by one level.

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -448,6 +448,79 @@ def test_program_filters(
...
+            'perm_filter': 'read',
+            'expected_programs': {
+                'masters-in-cs',
+                'masters-in-me',
+                'masters-in-philosophy',
+            },
+            'global_perm': True,
+        },
+        {
+            'groups': {},
```
kdmccormick: `set()`, not `{}`

kdmccormick (approved):
Looks good otherwise

registrar/apps/api/v1/tests/test_views.py
kdmccormick: (`{}` is empty dict)

schenedx: Done

PR #218, opened 2019-12-09
cleanup old program model

zacharis278:
@edx/masters-devs 

### See #214
Final migration to cleanup unused Program model.


schenedx (approved):
👍

PR #219, opened 2019-12-10
Refactor filestore to handle multiple buckets

kdmccormick:
## Description

When thinking about how to provide URLs to program reports, I realized that Registrar now needs to support reading from more than one bucket.

We currently use [Django Storages](https://django-storages.readthedocs.io/en/latest/) to interact with S3 using `S3Boto3Storage`, and we use its `FileSystemStorage` class for local development. That library supports multiple S3 buckets.

However, we wrap references to Django storages in an abstraction called `filestore`, which exists to handle some discrepancies with how download URLs are generated. `filestore` was designed with the assumption that there would only be one S3 bucket.

The aim of this PR is to make `filestore` work well with multiple buckets. Additionally, for local development, it would simulate having multiple buckets by putting each "bucket's" contents in a different folder.

## Meta
@edx/masters-devs 
Blocked by https://github.com/edx/edx-internal/pull/1348
Ticket: https://openedx.atlassian.net/browse/MST-58

### DIFF
--- a/registrar/apps/api/v1/tests/test_views.py
+++ b/registrar/apps/api/v1/tests/test_views.py
@@ -619,8 +619,8 @@ class ProgramCourseListViewTests(RegistrarAPITestCase, AuthRequestMixin):
     event = 'registrar.v1.get_program_courses'
 
     program_uuid = str(uuid.uuid4())
-    program_title = Faker().sentence(nb_words=6)  # pylint: disable=no-member
-    program_url = Faker().uri()  # pylint: disable=no-member
+    program_title = Faker().sentence(nb_words=6)
+    program_url = Faker().uri()
     program_type = 'Masters'
 
     @ddt.data(True, False)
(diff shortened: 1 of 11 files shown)


registrar/apps/core/filestore.py
```diff
@@ -119,19 +121,43 @@ class S3Filestore(FilestoreBase):
...
+    """
+    return get_filestore(settings.REGISTRAR_BUCKET, 'job-results')
+
+
+def get_program_reports_filestore():
+    """
+    Get filestore instance for program analytics reports.
+    """
+    # TODO: what is the correct path prefix for these reports?
+    return get_filestore(settings.PROGRAM_REPORTS_BUCKET, 'CHANGEME')
```
zacharis278: this isn't set on the output side yet so let's just pick something.  `reports` maybe?

zacharis278 (approved):
💃

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -198,7 +197,7 @@ def setUpClass(cls):
         cls._s3_mock = moto.mock_s3()
         cls._s3_mock.start()
         conn = boto3.resource('s3', region_name='us-west-1')
-        conn.create_bucket(Bucket=settings.AWS_STORAGE_BUCKET_NAME)
+        conn.create_bucket(Bucket=settings.REGISTRAR_BUCKET)
```
michaelroytman: The program reports list view will be going in the `api` app. How can the tests for that view leverage this mixin if the bucket being created is the Registrar bucket? Can we create a second bucket for the program reports?

kdmccormick: Good point. This mixin needs to handle both buckets. I'll revise this.

PR #221, opened 2019-12-11
Add read program reports role

how015:
MST-50:
In order to provide user access to reports, we need to create a new role for access to program reports.

@edx/masters-devs

### DIFF
--- a/registrar/apps/core/permissions.py
+++ b/registrar/apps/core/permissions.py
@@ -14,6 +14,9 @@ APP_PREFIX = 'core.'
 ORGANIZATION_READ_METADATA_KEY = 'organization_read_metadata'
 ORGANIZATION_READ_ENROLLMENTS_KEY = 'organization_read_enrollments'
 ORGANIZATION_WRITE_ENROLLMENTS_KEY = 'organization_write_enrollments'
+PROGRAM_READ_METADATA_KEY = 'program_read_metadata'
+PROGRAM_READ_REPORTS_KEY = 'program_read_reports'
+
 
 # A user with this permission can read any metadata
 # about an organization, including metadata about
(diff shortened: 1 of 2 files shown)


registrar/apps/core/permissions.py
```diff
@@ -33,13 +36,27 @@
 ORGANIZATION_WRITE_ENROLLMENTS = APP_PREFIX + ORGANIZATION_WRITE_ENROLLMENTS_KEY
 
 
+# A user with this permission can read any metadata about an program.
+PROGRAM_READ_METADATA = APP_PREFIX + PROGRAM_READ_METADATA_KEY
```
schenedx, line 40: Why do we need this permission, and don't have an associated role for this? Maybe we need a `ProgramMetadataRole`?

registrar/apps/core/permissions.py
```diff
@@ -107,6 +124,39 @@ class OrganizationReadWriteEnrollmentsRole(OrganizationRole):
     )
 
 
+class ProgramRole(object):
```
schenedx: This is a poor name for the base role class. How about `ProgramSpecificRoleBase`?

schenedx (commented):
See my inline comments

registrar/apps/core/permissions.py
```diff
@@ -33,13 +36,27 @@
 ORGANIZATION_WRITE_ENROLLMENTS = APP_PREFIX + ORGANIZATION_WRITE_ENROLLMENTS_KEY
 
 
+# A user with this permission can read any metadata about an program.
```
schenedx: #NIT `about a program`

schenedx (approved):
👍 LGTM. Good work

registrar/apps/core/permissions.py
```diff
@@ -107,6 +124,51 @@ class OrganizationReadWriteEnrollmentsRole(OrganizationRole):
     )
 
 
+class ProgramSpecificRoleBase(object):
```
zacharis278, line 127: Do we need to make this specific to program?  Since we want to the ability grant access to reports at the organization level as well we would also need a duplicate of this ReadReportsRole that works with an OrgGroup.  I'm thinking the base class for our roles can just be made generic to avoid this altogether.

zacharis278 (changes requested):

PR #222, opened 2019-12-12
Refactor role base to accomodate both org and program

how015:
MST-50
Refactor OrganazitionRole to make it work with both organization and program

@edx/masters-devs


kdmccormick (changes requested):
Going to request changes on a higher level here.

There shouldn't be a distinction between `ORGANIZATION_`-level and `PROGRAM_`-level permissions. I thought the reason we `Program` to `core` was so that we wouldn't have to make that distinction.

Permissions should be:
* `READ_METADATA`
* `READ_ENROLLMENTS`
* `WRITE_ENROLLMENTS`
* `READ_REPORTS`

Then, by being assigned to specific `Organizations` and `Programs` those permissions are give org- or program-level context.

cc'ing @Zacharis278 @MichaelRoytman in case they have a different take on this.

zacharis278:
@kdmccormick agreed, we chatted about this offline.  The goal is to remove the specific model context from these permissions.  They should match what you have listed here.

schenedx (approved):
👍

zacharis278 (approved):
:shipit:

registrar/apps/core/models.py
```diff
@@ -57,9 +57,9 @@ class Organization(TimeStampedModel):
     class Meta(object):
         app_label = 'core'
         permissions = (
-            (perms.ORGANIZATION_READ_METADATA_KEY, 'View Organization Metadata'),
-            (perms.ORGANIZATION_READ_ENROLLMENTS_KEY, 'Read Organization enrollment data'),
-            (perms.ORGANIZATION_WRITE_ENROLLMENTS_KEY, 'Write Organization enrollment data'),
+            (perms.READ_METADATA_KEY, 'View Organization Metadata'),
```
kdmccormick: `'View Metadata'`

registrar/apps/core/permissions.py
```diff
@@ -11,49 +11,33 @@
...
 # globally.
-ORGANIZATION_READ_METADATA_KEY = 'organization_read_metadata'
-ORGANIZATION_READ_ENROLLMENTS_KEY = 'organization_read_enrollments'
-ORGANIZATION_WRITE_ENROLLMENTS_KEY = 'organization_write_enrollments'
-PROGRAM_READ_METADATA_KEY = 'program_read_metadata'
-PROGRAM_READ_REPORTS_KEY = 'program_read_reports'
+READ_METADATA_KEY = 'read_metadata'
+READ_ENROLLMENTS_KEY = 'read_enrollments'
+WRITE_ENROLLMENTS_KEY = 'write_enrollments'
+READ_REPORTS_KEY = 'read_reports'
```
kdmccormick, line 17: As @Zacharis278 @how015 and I talked about offline, we unfortunately need to keep the old codenames (with the `organization_`) prefix for now. Otherwise, everybody who had the `organization_*` permissions would lose them when this PR was deployed.

registrar/apps/core/permissions.py
```diff
@@ -64,113 +48,78 @@
...
 
-class OrganizationReadMetadataRole(OrganizationRole):
+class ReadMetadataRole(RoleBase):
     """
-    The least-privileged organization-scoped role, it may only
-    read metadata about an organization's programs and courses.
+    The least-privileged role, it may only read metadata about an organization or a program.
     """
-    name = 'organization_read_metadata'
+    name = 'read_metadata'
```
kdmccormick: Similarly, this `name` needs to stay prefixed with `organization_`. But the class name and all the other refactors are still good.

registrar/apps/core/migrations/0007_auto_20191216_1518.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
```
kdmccormick: (per the other changes, remove this and roll back locally)

kdmccormick (changes requested):
Comments that we talked about. Great refactoring though!

PR #223, opened 2019-12-13
Pin the library django-model-utils to older version

schenedx:
The new version for library django-model-utils 4.0.0 requires django 2.0 or above. This causes failure when trying to upgrade request of the libraries requirements. Fix this by pinning django-model-utils to version less than 4.0.0

@edx/masters-devs

### DIFF
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -10,3 +10,4 @@
 
 # These packages are pinned because their next version would require django 2.0.0
 django-guardian==2.0.0
+django-model-utils<4.0.0


requirements/constraints.txt
```diff
@@ -9,4 +9,5 @@
 # linking to it here is good.
 
 # These packages are pinned because their next version would require django 2.0.0
-django-guardian==2.0.0
\ No newline at end of file
+django-guardian==2.0.0
+django-model-utils<=3.2.0
```
kdmccormick: The minimally restrictive constraint would be `django-model-utils<4.0.0`.

kdmccormick (approved):
One suggestion, but looks good nonetheless.

requirements/constraints.txt
schenedx: Agree. Done.

PR #225, opened 2019-12-17
Add logging when filestore operations fail

kdmccormick:
#219 is bugged. If you try to download or upload enrollments on Stage, the job fails. Looking at Splunk, an AccessDenied is happening when trying to save to S3. I'm thinking that the bucket name might be misconfigured.

This PR adds logging to help diagnose it.

**Bonus:** Fixes our only RemoveInDjango20Warning

### DIFF
--- a/registrar/apps/core/models.py
+++ b/registrar/apps/core/models.py
@@ -122,6 +122,8 @@ class OrganizationGroup(Group):
 
     .. no_pii::
     """
+    objects = models.Manager()
+
     class Meta(object):
         app_label = 'core'
         verbose_name = 'Organization Group'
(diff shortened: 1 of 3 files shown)


registrar/apps/core/models.py
```diff
@@ -122,6 +122,8 @@ class OrganizationGroup(Group):
 
     .. no_pii::
     """
+    objects = models.Manager()
```
michaelroytman, line 125: Is this related to the logging?

michaelroytman (approved):

registrar/apps/core/models.py
kdmccormick: Nah, it's to silence a RemovedInDjango20Warning. It's in its own commit.

PR #226, opened 2019-12-18
Add a ReportsListView.

michaelroytman:
## Description

Add a `ReportsListView` to the `api` Django app for surfacing metadata information about reports available for a program.

See https://openedx.atlassian.net/browse/MST-57.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 python:
-    - 3.6
+    - 3.5.2
 sudo: false
 
 services:
--- a/api.yaml
+++ b/api.yaml
@@ -407,6 +407,33 @@ endpoints:
         description: >
           Course does not exist within program, or program was not found.
 
+  list_program_reports: &list_program_reports
+    summary: List program reports
+    description: List the reports in a program
+    parameters:
+      - *program_path_param
+      - in: query
+        name: min_created_date
+        type: date
+        description: >
+          The ISO-format date used to limit the reports
+          returned based on their created date
+        required: false
+        example: '2019-12-18'
+
+
+    responses:
+      200:
+        description: OK
+        schema:
+          type: array
+          items:
+            $ref: '#/models/ProgramReport'
+      403:
+        description: User is not authorized to view program reports.
+      404:
+        description: Program was not found.
+
 
   swagger:
     get:
@@ -507,6 +534,12 @@ paths:
       <<: *patch_program_enrollments
       operationId: registrar_v1_patch_program_enrollments
       tags: ['v1']
+  '/v1/programs/{program_key}/reports':
+    get:
+      <<: *list_program_reports
+      operationId: registrar_v1_list_program_reports
+      tags: ['v1']
+
 
   ### V2 API ###
   '/v2/jobs/{job_id}/':
@@ -559,6 +592,11 @@ paths:
     patch:
       <<: *patch_program_enrollments
       operationId: registrar_v2_patch_program_enrollments
+      tags: ['v2']
+  '/v2/programs/{program_key}/reports':
+    get:
+      <<: *list_program_reports
+      operationId: registrar_v2_list_program_reports
       tags: ['v2']
 
 ################################################################################
@@ -631,3 +669,18 @@ models:
         format: uri
         nullable: true
         description: URL to result file; null if result unavailable
+  ProgramReport:
+    type: object
+    properties:
+      name:
+        type: string
+        description: The name of the program report
+      created_date:
+        type: string
+        format: date
+        description: >
+          The ISO-format date-time that the program report was created
+      download_url:
+        type: string
+        format: uri
+        description: URL to download the program report
--- a/registrar/apps/api/v1/tests/test_views.py
+++ b/registrar/apps/api/v1/tests/test_views.py
@@ -16,7 +16,6 @@ from django.conf import settings
 from django.core.cache import cache
 from django.urls import reverse
 from faker import Faker
-from freezegun import freeze_time
 from guardian.shortcuts import assign_perm
 from rest_framework.test import APITestCase
 from user_tasks.models import UserTaskStatus
@@ -55,6 +54,7 @@ from registrar.apps.core.tests.factories import (
     ProgramOrganizationGroupFactory,
     UserFactory,
 )
+from registrar.apps.core.tests.freezegun_wrapper import freeze_time
 from registrar.apps.core.tests.utils import mock_oauth_login
 from registrar.apps.core.utils import serialize_to_csv
 from registrar.apps.enrollments.data import (
--- a/registrar/apps/api/v1/tests/test_views.py
+++ b/registrar/apps/api/v1/tests/test_views.py
@@ -16,6 +16,7 @@ from django.conf import settings
 from django.core.cache import cache
 from django.urls import reverse
 from faker import Faker
+from freezegun import freeze_time
 from guardian.shortcuts import assign_perm
 from rest_framework.test import APITestCase
 from user_tasks.models import UserTaskStatus
@@ -29,8 +30,12 @@ from registrar.apps.api.v1.views import (
 )
 from registrar.apps.common.constants import PROGRAM_CACHE_KEY_TPL
 from registrar.apps.common.data import DiscoveryCourseRun, DiscoveryProgram
+from registrar.apps.common.tests.mixins import S3MockEnvVarsMixin
 from registrar.apps.core import permissions as perms
-from registrar.apps.core.filestore import get_enrollment_uploads_filestore
+from registrar.apps.core.filestore import (
+    get_enrollment_uploads_filestore,
+    get_program_reports_filestore,
+)
 from registrar.apps.core.jobs import (
     post_job_failure,
     post_job_success,
@@ -136,7 +141,13 @@ class RegistrarAPITestCase(TrackTestMixin, APITestCase):
             organization=cls.hum_org,
             role=perms.OrganizationReadMetadataRole.name
         )
+        cls.hum_data_op_group = OrganizationGroupFactory(
+            name='hum-data-ops',
+            organization=cls.hum_org,
+            role=perms.OrganizationReadReportRole.name
+        )
         cls.hum_admin.groups.add(cls.hum_admin_group)  # pylint: disable=no-member
+        cls.hum_admin.groups.add(cls.hum_data_op_group)  # pylint: disable=no-member
 
         cls.program_user = UserFactory(username='program-admin')
         cls.program_group = ProgramOrganizationGroupFactory(
@@ -196,7 +207,7 @@ class RegistrarAPITestCase(TrackTestMixin, APITestCase):
         )
 
 
-class S3MockMixin(object):
+class S3MockMixin(S3MockEnvVarsMixin):
     """
     Mixin for classes that need to access S3 resources.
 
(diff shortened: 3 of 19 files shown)


registrar/apps/api/v1/tests/test_views.py
```diff
@@ -198,6 +208,11 @@ class S3MockMixin(object):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        # required to prevent moto from mutating real infrastructure
+        os.environ['AWS_ACCESS_KEY_ID'] = 'testing'
```
michaelroytman: Is there a better way to do this?

michaelroytman: See [here](https://github.com/spulec/moto#how-do-i-avoid-tests-from-mutating-my-real-infrastructure) for why this is necessary.

registrar/apps/api/v1/views.py
```diff
@@ -552,3 +556,123 @@ def get(self, request, *args, **kwargs):
...
+                'name': report_name,
+                'created_date': self._get_file_created_date(
+                    file_path, filename_date_format_string,
+                    output_date_format_string
+                ),
+                'download_url': None,
+            }
+            reports_metadata.append(report_metadata)
+
+        if 'created_date_on_or_after' in request.query_params:
```
michaelroytman: I would **love** an idea for a better name.

registrar/apps/api/v1/views.py
```diff
@@ -552,3 +556,123 @@ def get(self, request, *args, **kwargs):
...
+
+        Parameters:
+            - filename: the name of the file
+
+        Returns:
+            - String: the date the file was created as a YYYY-MM-DD formatted string
+            - None: if the date is not in the filename or the date is misformatted
+        """
+        # pull out the date string from the filename
+        pattern = re.compile(r'.*__(\d*_\d*_\d*)[.]*\w*')
```
michaelroytman, line 650: I'm not sure whether all this error checking is necessary. The data pipeline writes the files with the date in the filenames. I tend to err on the side of caution, but I'm happy to hear arguments for removing it.

api.yaml
```diff
@@ -407,6 +407,31 @@ endpoints:
...
+    responses:
+      200:
+        description: OK
+        schema:
+          type: array
+          items:
+            $ref: '#/models/ProgramReport'
+      403:
+        description: User is not authorized to view program reports.
+
```
kdmccormick, line 436: Might be worth adding the `404` response for `program not found` and any other 404 scenarios.

registrar/apps/api/v1/tests/test_views.py
kdmccormick: Interesting. `test_filestore.py` and `test_tasks.py` also should have these set probably (they don't share `S3MockMixin` because of some weirdness with how moto is sensitive to imports; I don't remember the details, but I couldn't get it to work).

For the medium-term in this PR, I'd recommend a separate simple mixin that just sets those environment variables, and use that mixin for all tests that use moto.

In the long-term, we have https://openedx.atlassian.net/browse/EDUCATOR-4773 to remove moto.

registrar/apps/api/v1/views.py
kdmccormick: `created_on_or_after`? (I think it's still clear that it's a date)

Alternatively, `min_date_created` / `min_created_date`?

registrar/apps/api/v1/views.py
kdmccormick: I like it! In the spirit of decoupled microservices, I think it's best not to assume that other services are working the way we expect them to.

registrar/apps/api/v1/views.py
```diff
@@ -552,3 +556,123 @@ def get(self, request, *args, **kwargs):
...
+            self._log_invalid_filename(filename)
+            return None
+
+        date_string = match.group(1)
+
+        try:
+            # validate that the date is actually a date; otherwise,
+            # return None
+            date = datetime.strptime(date_string, filename_date_format_string)
+            return datetime.strftime(date, output_date_format_string)
```
kdmccormick: Nit: The `return datetime.strftime(...` belongs outside the try-catch, because if that line fails, it's because the output format is bad, not because the date string from the filename is bad.

registrar/apps/api/v1/views.py
kdmccormick: Also, I appreciate the helpful logging messages :+1:

registrar/apps/api/v1/views.py
```diff
@@ -552,3 +556,123 @@ def get(self, request, *args, **kwargs):
...
+        except ValueError:
+            self._log_invalid_filename(filename)
+            return None
+
+    def _log_invalid_filename(self, filename):
+        logger.warning('Filename {} does not contain a valid date.'.format(filename))
```
kdmccormick: Nit: Consider rephrasing to something like `is not in the expected format: uuid__YYYY_MM_DD.extension`, because the filename might contain a valid-but-incorrectly-formatted date like `2000-01-01`, which would trigger this log statement.

registrar/apps/api/v1/views.py
```diff
@@ -552,3 +556,123 @@ def get(self, request, *args, **kwargs):
...
+        try:
+            reports = filestore.list(file_prefix)
+        except FileNotFoundError:
+            raise Http404()
+
+        # list method of the filestore returns a 2-tuple containing a
+        # list of directories and a list of files on the path, respectively,
+        # so iterate the list of files
+        for report_name in reports[1]:
+            file_path = filestore.get_full_path('{}/{}'.format(file_prefix, report_name))
```
kdmccormick: `report_name` would be sufficient to pass to `_get_file_created_date`, right?

registrar/apps/core/tests/test_filestore.py
```diff
@@ -68,6 +69,14 @@ def test_s3_filestore(self, bucket, location, prefix, path, contents):
             url = filestore.store(path, contents)
             self.assertTrue(filestore.exists(path))
             self.assertIn(location, url)
+
+            pure_path = PurePath(path)
+            parent_path = pure_path.parent.name
+            file_name = pure_path.name
+
+            files = filestore.list(parent_path)
```
kdmccormick: Nit: What you're assigning to `files` is more like `files_and_dirs`.

how about `files = filestore.list(parent_path)[1]` ?

registrar/apps/core/tests/test_filestore.py
```diff
@@ -68,6 +69,14 @@ def test_s3_filestore(self, bucket, location, prefix, path, contents):
...
+            self.assertEqual(files[1], [file_name] if file_name else [])
```
kdmccormick: When would `file_name` fail the if-check?

registrar/apps/api/v1/views.py
```diff
@@ -552,3 +556,123 @@ def get(self, request, *args, **kwargs):
...
+        filestore = get_program_reports_filestore()
+        file_prefix = '{}/{}'.format(self.program.managing_organization.key, self.program.discovery_uuid)
+        filename_date_format_string = '%Y_%m_%d'
+        output_date_format_string = '%Y-%m-%d'
+
+        reports_metadata = []
+
+        try:
+            reports = filestore.list(file_prefix)
+        except FileNotFoundError:
```
michaelroytman: Django's `FileStorageBackend`'s implementation of this method returns `FileNotFoundError` when a file is not found. However, `S3Boto3Storage`'s implementation of this method does not; it just returns a 2-tuple of empty lists. It's annoying to have production code respect implementation details of a backend that is only used in a development environment. I could wrap the `FileStorageBackend`'s call to `list_dir` with a try-except and return a 2-tuple of empty lists, like the `S3Boto3Storage` does. Thoughts?

registrar/apps/api/v1/views.py
```diff
@@ -552,3 +556,123 @@ def get(self, request, *args, **kwargs):
...
+        # so iterate the list of files
+        for report_name in reports[1]:
+            file_path = filestore.get_full_path('{}/{}'.format(file_prefix, report_name))
+            report_metadata = {
+                'name': report_name,
+                'created_date': self._get_file_created_date(
+                    file_path, filename_date_format_string,
+                    output_date_format_string
+                ),
+                'download_url': None,
```
kdmccormick: With my PR merged, you should be able to do
```python
            'download_url': filestore.get_url('{}/{}'.format(file_prefix, report_name))
```

registrar/apps/api/v1/views.py
```diff
@@ -552,3 +556,123 @@ def get(self, request, *args, **kwargs):
...
         )
+
+
+class ReportsListView(ProgramSpecificViewMixin, APIView):
+    """
+    A view for listing metadata about reports for a program.
+
+    Path: /api/[version]/programs/{program_key}/reports?created_date_on_or_after={created_date_on_or_after}
+
+    Example Response:
```
kdmccormick, line 557: Nice examples!

registrar/apps/api/v1/views.py
```diff
@@ -552,3 +556,123 @@ def get(self, request, *args, **kwargs):
...
+        },
+        {
+            "name":"individual_report__2019_12_12.txt",
+            "created_date":"2019_12_12",
+            "download_url":null
+        }
+    ]
+
+    Returns:
+     * 202: Returns a list of metadata about available reports for a program.
```
kdmccormick: `200`

kdmccormick (commented):
Looks great overall; I have a few comments, but all minor.

registrar/apps/api/v1/views.py
kdmccormick: I like your idea a lot. The idea of `filestore` is to smooth over the implementation discrepancies of `FileSystemStorage` vs. `S3Boto3Storage`, so I think it would fit well there.

registrar/apps/api/v1/views.py
```diff
@@ -552,3 +556,123 @@ def get(self, request, *args, **kwargs):
...
+        file_prefix = '{}/{}'.format(self.program.managing_organization.key, self.program.discovery_uuid)
+        filename_date_format_string = '%Y_%m_%d'
+        output_date_format_string = '%Y-%m-%d'
+
+        reports_metadata = []
+
+        try:
+            reports = filestore.list(file_prefix)
+        except FileNotFoundError:
+            raise Http404()
```
zacharis278: If the program itself is valid and we're asking for a collection of items I think this should just be returning an empty array. (sounds like you may have that solved in the solution above anyway?)

registrar/apps/api/v1/views.py
zacharis278: +1 for `min_created_date`

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: python
 python:
-    - 3.6
+    - 3.5.2
```
michaelroytman, line 3: This is to match the version of Python we use in development.

registrar/apps/api/v1/views.py
```diff
@@ -542,3 +546,121 @@ def get(self, request, *args, **kwargs):
...
+        reports = filestore.list(file_prefix)[1]
+
+        for report_name in reports:
+            report_metadata = {
+                'name': report_name,
+                'created_date': self._get_file_created_date(
+                    report_name, filename_date_format_string,
+                    output_date_format_string
+                ),
+                'download_url': filestore.get_url('{}/{}'.format(file_prefix, report_name))
```
zacharis278: This may be a premature optimization but I'm thinking we should do this url generation after filtering down the list of reports?  As is we're calling out to AWS to generate a signed url for every file in that folder every time the UI is loaded.

zacharis278 (approved):
Added one non-blocking comment about optimizing our url generation.  LGTM

.travis.yml
kdmccormick: Nice.

kdmccormick (approved):
+1 to @Zacharis278 's suggested URL optimization.

Otherwise, looks great!

registrar/apps/core/tests/test_filestore.py
```diff
@@ -69,6 +71,14 @@ def test_s3_filestore(self, bucket, location, prefix, path, contents):
             url = filestore.store(path, contents)
             self.assertTrue(filestore.exists(path))
             self.assertIn(location, url)
+
+            pure_path = PurePath(path)
```
kdmccormick, line 75: I hadn't heard of `pathlib`... definitely going to try to use it going forward.

registrar/apps/api/v1/views.py
kdmccormick: @MichaelRoytman Did you end up coming to a conclusion on this?

registrar/apps/api/v1/views.py
michaelroytman: Ah, good point! I'll make that change.

PR #228, opened 2019-12-18
Add ProgramOrgGroup

how015:
MST-60

Create a ProgramOrgGroup model. This model will have a foreign key field to a Program, a foreign key to an Organization, and a role field. The reason we include the Organization in this model is to contextualize the permissions of a program within an organization.

This addresses the problem of a Program A, authored jointly by Organizations 1 and 2. If User X is given program level report access to Program A, they would get access to two nearly identical reports - one for Organization 1 and one for Organization 2, where the two reports differ only in the "authoring organization" column. This avoids this prob

Add sufficient checks/constraints so that a ProgramOrgGroup cannot have an Organization value that isn’t an authoring org of the program

Update permission checking logic to support multiple permission objects (e.g Org or Program)

mixins.py#L80


codecov.yml
```diff
@@ -9,3 +9,4 @@ coverage:
         target: 100
 ignore:
   - "registrar/apps/core/permissions.py"
+  - "registrar/apps/core/models.py"
```
zacharis278: why do we need to ignore test coverage on this file?

registrar/apps/api/v1/mixins.py
```diff
@@ -99,21 +99,27 @@ def check_permissions(self, request):
...
             }
         else:
-            missing_permissions = missing_global_permissions
+            missing_permissions = missing_global_permissions  # pragma: no cover
 
         if missing_permissions:
             self.add_tracking_data(missing_permissions=list(missing_permissions))
             self._unauthorized_response()
 
+    def _has_permission(self, user, perm, objects):
```
kdmccormick: Nice use of a helper function to simplify code.

Consider naming it `_has_permission_on_any` to clarify that permission on **any** object is sufficient, as opposed to requiring it on **all** objects, which might be assumed by a reader.

Also, you can make it more clear that this function doesn't reference or mutate the `self` object by making it static:
```python
@staticmethod
def _has_permission_on_any(user, perm, objects):
    ...
```

Note that you call static functions in Python with `self.`, just like any other method.

registrar/apps/api/v1/mixins.py
```diff
@@ -52,7 +52,7 @@ class AuthMixin(TrackViewMixin):
     raise_404_if_unauthorized = False
     staff_only = False
 
-    def get_permission_object(self):
+    def get_permission_objects(self):
         """
         Get an object against which required permissions will be checked.
```
kdmccormick: Update the docstring to reflect the new behavior of this method: that it returns a list of objects, for which permission on at least one is required.

registrar/apps/api/v1/mixins.py
```diff
@@ -52,7 +52,7 @@ class AuthMixin(TrackViewMixin):
...
```
kdmccormick, line 58: I recommend that we return `[]` (empty list) instead of `None` from this method by default.

`[]` makes sense semantically ("there are no objects on which can have permission, so global permissions are required"), and it would allow us to remove the `elif objects:` branch on line 105, because calling `self._has_permission` with an empty list will always return `False`.

registrar/apps/core/permissions.py
```diff
@@ -121,5 +138,6 @@ class ReadReportRole(RoleBase):
     ReadMetadataRole,
     ReadEnrollmentsRole,
     ReadWriteEnrollmentsRole,
-    ReadReportRole
+    ReadReportRole,
+    ProgramReadMetadataRole
```
kdmccormick: Nit: Prefer trailing commas, as they make diffs nicer and make it easier to add new items to the list.

registrar/apps/core/models.py
```diff
@@ -172,6 +175,64 @@ def __str__(self):
         return 'OrganizationGroup: {} role={}'.format(self.organization.name, self.role)
 
 
+class ProgramOrgGroup(Group):
```
kdmccormick: Consider naming this `ProgramOrganizationGroup` for the sake of consistency with other classes which spell out the word Organization.

registrar/apps/core/models.py
```diff
@@ -172,6 +175,64 @@ def __str__(self):
...
+        max_length=255,
+        choices=ROLE_CHOICES,
+        default=perms.ReadMetadataRole.name,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        try:
+            self._initial_program = self.program
+            self._initial_organization = self.program.managing_organization
```
kdmccormick: You can nix the `_initial_organization` member.

It exists on `OrganizationGroup` so that we can un-assign permissions from an org when an org group is changed to a different org.

Because this class operates on Programs, we just need to un-assign permissions from the old program when an instance of it is changed to a different program, which you've already taken care of in the `save` method you wrote. So, there's no need for the `_initial_organization` member.

registrar/apps/core/models.py
```diff
@@ -172,6 +175,64 @@ def __str__(self):
         return 'OrganizationGroup: {} role={}'.format(self.organization.name, self.role)
 
 
+class ProgramOrgGroup(Group):
+    """
+    Group subclass to grant select guardian permissions to a group on a program level.
+
+    .. no_pii::
+    """
+
```
kdmccormick, line 186: The majority of the functions and fields declared in this class are duplicated with those in `OrganizationGroup`.

I recommend doing something like this:
```python

class OrganizationGroupMixin(object):
    # shared fields & methods

class OrganizationGroup(OrganizationGroupMixin, Group):
   # org-specific stuff

class ProgramOrganizationGroup(OrganizationGroupMixin, Group):
   # program-specific stuff
```

registrar/apps/core/permissions.py
```diff
@@ -33,11 +35,15 @@
 READ_REPORTS = APP_PREFIX + READ_REPORTS_KEY
 
 
+PROGRAM_READ_METADATA = APP_PREFIX + PROGRAM_READ_METADATA_KEY
```
kdmccormick, line 42: I'm a bit confused about why we need a program-read-metadata permission but not any other program-specific permissions. Let's chat.

kdmccormick (commented):
Good work! Some comments here; also, let's chat about the program vs. org role/permission duplication thing.

registrar/apps/api/v1/views.py
```diff
@@ -85,17 +85,26 @@ class ProgramListView(AuthMixin, TrackViewMixin, ListAPIView):
...
+                if user.has_perm(perms.PROGRAM_READ_METADATA, program)
+            )
             return programs
 
         # if the user has permissions across organizations
         # via membership in a "global-access" group, give them
         # access to all programs that fits their permission criteria
-        user = self.request.user
+
         if not user.has_perm(self.permission_filter):
```
zacharis278: This section of code won't return programs if the user only has the requested permission at the program level.  This needs to check the program object with the appropriate permission as well.  PERMISSION_QUERY_PARAM map would need to be updated to support this.

registrar/apps/api/constants.py
```diff
@@ -10,12 +10,12 @@
...
 PERMISSION_QUERY_PARAM_MAP = {
-    'metadata': perms.READ_METADATA,
-    'read': perms.READ_ENROLLMENTS,
-    'write': perms.WRITE_ENROLLMENTS,
+    'metadata': perms.ORGANIZATION_READ_METADATA,
+    'read': perms.ORGANIZATION_READ_ENROLLMENTS,
+    'write': perms.ORGANIZATION_WRITE_ENROLLMENTS,
 }
 
 ENROLLMENT_PERMISSIONS_LIST = [
```
zacharis278, line 18: I think this should also include the program permissions

registrar/apps/core/permissions.py
```diff
@@ -68,58 +97,152 @@ def assign_to_group(cls, group, model):
...
+        PROGRAM_READ_REPORTS,
     )
 
 
-ROLES = [
-    ReadMetadataRole,
-    ReadEnrollmentsRole,
-    ReadWriteEnrollmentsRole,
-    ReadReportRole
+class APIPermissionBase(object):
```
zacharis278, line 198: nice, I like this solution 👍

registrar/apps/api/mixins.py
```diff
@@ -107,5 +108,12 @@ def _track(self, status_code):
             '%s invoked on Registrar by user with ID=%s with properties %s',
             event_name,
             self.request.user.id,
-            json.dumps(properties, skipkeys=True, sort_keys=True),
+            json.dumps(properties, skipkeys=True, sort_keys=True, cls=CustomEncoder),
         )
+
+
+class CustomEncoder(json.JSONEncoder):
```
zacharis278, line 115: Can you add a docstring explaining what this does?

registrar/apps/core/models.py
```diff
@@ -172,6 +177,64 @@ def __str__(self):
...
+        app_label = 'core'
+        verbose_name = 'Program Group'
+
+    ROLE_CHOICES = (
+        (role.name, role.description)
+        for role in perms.PROGRAM_ROLES
+    )
+
+    program = models.ForeignKey(Program, on_delete=models.CASCADE)
+    organization = models.ForeignKey(Organization, on_delete=models.CASCADE)
```
zacharis278: can we change this to something like `granting_organization`.  Hopefully that will it more clear to users using the admin page that Program is the model being assigned a role.

registrar/apps/core/permissions.py
```diff
@@ -7,37 +7,66 @@
...
+
+# A user with this permission can write any enrollment data contained in a program.
+PROGRAM_WRITE_ENROLLMENTS = APP_PREFIX + PROGRAM_WRITE_ENROLLMENTS_KEY
+
+
+# A user with this permission can read any reports of a program.
+PROGRAM_READ_REPORTS = APP_PREFIX + PROGRAM_READ_REPORTS_KEY
 
 
 PERMISSIONS = {
```
zacharis278: if this is just Organization permissions can we refactor the name to make that more clear?

registrar/apps/core/models.py
how015: will do this in a separate PR because this change requires data migration, I tend to do it safely :) https://openedx.atlassian.net/browse/MST-87

registrar/apps/core/models.py
how015: chatted offline with Kyle, we decided not to change here.

zacharis278 (changes requested):
Overall looks great just a lot of little comments.  I do think we need to add several tests to cover the case of having permissions assigned at the ProgramOrgGroup on the models and for each affected endpoint.

registrar/apps/core/models.py
how015: Agree with you, Kyle, but per MST-60's task description, "The reason we include the Organization in this model is to contextualize the permissions of a program within an organization."

registrar/apps/core/models.py
zacharis278: If this PR is creating that model in the first place that original migration can be recreated now before committing the model to master.  Let me know if you wanna go over this and we can walk through it in person. You'll be much better off doing that because changing model names after the fact is cumbersome due to our blue-green deployments.

registrar/apps/core/models.py
how015: https://openedx.atlassian.net/browse/MST-87 since this will result in data migration.

codecov.yml
how015: Will address it in https://openedx.atlassian.net/browse/MST-87

registrar/apps/core/permissions.py
kdmccormick: I balked at this when you first showed me it, but now that I'm looking at it more closely, I like the look of it. :+1:

registrar/apps/core/models.py
how015: got it, will do it in this PR

registrar/apps/api/v1/views.py
how015: right, fixed this logic like below

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -284,34 +293,25 @@ def test_all_programs_200(self):
             ]
         )
 
-    @ddt.data(
```
zacharis278, line 287: don't we still need these test cases?  Additionally it looks like the test_program_filters test doesn't yet cover the case of a user with permissions assigned at the program

registrar/apps/core/tests/test_models.py
```diff
@@ -68,9 +68,9 @@ def setUp(self):
         self.user = UserFactory()
```
zacharis278, line 72: do we have any tests for the new ProgramOrganizationGroup model?

zacharis278 (commented):
looks good.  Just a few comments on our test coverage

zacharis278 (approved):
I suggest leaving the strings documenting organization permissions (e.g. View organization metadata) alone just to avoid migrating that field and having to manually re-assign those options in prod.

Otherwise LGTM 💃

PR #233, opened 2020-01-14
Expose Program Permissions

zacharis278:
@edx/masters-devs 

- Adds a list of user APIPermissions to Program get/list endpoints
- `user_has_perm` query parameters now accept an APIPermission string.
- See [ADR on API Permissions](https://github.com/edx/registrar/blob/master/docs/decisions/0006-api-permissions.mdn)

### DIFF
--- a/registrar/apps/api/serializers.py
+++ b/registrar/apps/api/serializers.py
@@ -38,7 +38,7 @@ class ProgramSerializer(serializers.ModelSerializer):
             ).union(
                 get_user_api_permissions(user, program.managing_organization)
             )
-            return [permission.name for permission in user_permissions]
+            return sorted([permission.name for permission in user_permissions])
         else:
             return []
 
--- a/registrar/apps/api/serializers.py
+++ b/registrar/apps/api/serializers.py
@@ -4,10 +4,13 @@ should be created here. As the API evolves, serializers may become more
 specific to a particular version of the API. In this case, the serializers
 in question should be moved to versioned sub-package.
 """
+from guardian.shortcuts import get_perms
 from rest_framework import serializers
+from rest_framework.fields import CurrentUserDefault
 from user_tasks.models import UserTaskStatus
 
 from registrar.apps.core.models import Program
+from registrar.apps.core.utils import get_user_api_permissions
 from registrar.apps.enrollments.constants import (
     COURSE_ENROLLMENT_STATUSES,
     PROGRAM_ENROLLMENT_STATUSES,
@@ -22,10 +25,20 @@ class ProgramSerializer(serializers.ModelSerializer):
     program_key = serializers.CharField(source='key')
     program_title = serializers.CharField(source='title')
     program_url = serializers.URLField(source='url')
+    permissions = serializers.SerializerMethodField(source='get_permissions')
 
     class Meta:
         model = Program
-        fields = ('program_key', 'program_title', 'program_url', 'program_type')
+        fields = ('program_key', 'program_title', 'program_url', 'program_type', 'permissions')
+
+    def get_permissions(self, program):
+        request = self.context.get('request')
+        if request and hasattr(request, 'user'):
+            user = request.user
+            user_permissions = get_user_api_permissions(user, program).union(get_user_api_permissions(user, program.managing_organization))
+            return [permission.name for permission in user_permissions]
+        else:
+            return []
 
 
 class ProgramEnrollmentRequestSerializer(serializers.Serializer):
--- a/registrar/apps/api/serializers.py
+++ b/registrar/apps/api/serializers.py
@@ -4,9 +4,7 @@ should be created here. As the API evolves, serializers may become more
 specific to a particular version of the API. In this case, the serializers
 in question should be moved to versioned sub-package.
 """
-from guardian.shortcuts import get_perms
 from rest_framework import serializers
-from rest_framework.fields import CurrentUserDefault
 from user_tasks.models import UserTaskStatus
 
 from registrar.apps.core.models import Program
@@ -35,7 +33,11 @@ class ProgramSerializer(serializers.ModelSerializer):
         request = self.context.get('request')
         if request and hasattr(request, 'user'):
             user = request.user
-            user_permissions = get_user_api_permissions(user, program).union(get_user_api_permissions(user, program.managing_organization))
+            user_permissions = get_user_api_permissions(
+                user, program
+            ).union(
+                get_user_api_permissions(user, program.managing_organization)
+            )
             return [permission.name for permission in user_permissions]
         else:
             return []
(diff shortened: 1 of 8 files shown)


registrar/apps/api/serializers.py
```diff
@@ -22,10 +23,24 @@ class ProgramSerializer(serializers.ModelSerializer):
...
+    def get_permissions(self, program):
+        request = self.context.get('request')
+        if request and hasattr(request, 'user'):
+            user = request.user
+            user_permissions = get_user_api_permissions(
+                user, program
+            ).union(
+                get_user_api_permissions(user, program.managing_organization)
+            )
+            return sorted([permission.name for permission in user_permissions])
```
kdmccormick: Style nit: You can drop the square brackets, as `permission.name for permission in user_permissions` is a generator which is fine to pass directly to `sorted`.

registrar/apps/api/v1/views.py
```diff
@@ -162,11 +162,17 @@ def permission_filter(self):
...
-        else:
-            self.add_tracking_data(failure='no_such_perm')
-            raise Http404()
+
+        try:
+            return next(p for p in perms.API_PERMISSIONS if p.name == perm_query_param)
+        except StopIteration:
+            # maintains functionality with the currently deployed version of the UI
+            # and can be removed once these query params are no loger in use
+            if perm_query_param in PERMISSION_QUERY_PARAM_MAP:
```
kdmccormick: Consider renaming `PERMISSION_QUERY_PARAM_MAP` to something that shows it's deprecated and only still in the code for backwards compatibility

registrar/apps/core/utils.py
```diff
@@ -24,6 +27,33 @@ def get_user_organizations(user):
...
+    Returns a set of all APIPermissions granted to the user on a
+    provided object instance. This includes permissions granted though a
+    global permission or role.
+    """
+    user_object_permissions = get_perms(user, obj)
+    user_global_permissions = list(user.user_permissions.all().values_list('codename', flat=True))
+
+    user_api_permissions = set()
+
+    api_permission_map = {}
```
kdmccormick: This logic computes the same dict every time this function is called. Consider just making it a top-level constant in `permissions.py`:
```python
def _build_db_to_api_permissions():
    """
    Returns a dict from database permission string (without app name) to
    corresponding APIPermissionBase instance.
    """
    # ... your logic

DB_TO_API_PERMISSIONS = _build_db_to_api_permissions()
```
In addition to re-usability, it'd be easier to debug the case where `DB_TO_API_PERMISSIONS` ends up with a nonsense value (as opposed to the bug being buried in `get_user_api_permissions`).

registrar/apps/core/utils.py
```diff
@@ -24,6 +27,33 @@ def get_user_organizations(user):
     return user_organizations
 
 
+def get_user_api_permissions(user, obj):
```
kdmccormick: Nice! I'm glad all this logic is collected here instead of being scattered throughout views.

registrar/apps/api/v1/views.py
```diff
@@ -162,11 +162,17 @@ def permission_filter(self):
         perm_query_param = self.request.GET.get('user_has_perm', None)
         if not perm_query_param:
             return None
-        elif perm_query_param in PERMISSION_QUERY_PARAM_MAP:
```
kdmccormick, line 169: Can you update the docstring to say that this returns an `APIPermissionBase` subclass, not a list of user permissions?

kdmccormick (commented):
Here's my first pass.

Also, I think we can simplify `ProgramListView.get_queryset` given the utilities that you wrote; I'll circle back to that thought later.

registrar/apps/core/utils.py
kdmccormick: One suggested refactor: allow `obj` to be passed in as `None`, and handle that as a global permissions check, which would allow this function to be used for all API permission checking.

registrar/apps/core/utils.py
```diff
@@ -27,29 +26,20 @@ def get_user_organizations(user):
...
 
-def get_user_api_permissions(user, obj):
+def get_user_api_permissions(user, obj=None):
     """
     Returns a set of all APIPermissions granted to the user on a
     provided object instance. This includes permissions granted though a
     global permission or role.
     """
-    user_object_permissions = get_perms(user, obj)
+    user_object_permissions = get_perms(user, obj) if obj is not None else []
```
kdmccormick, line 36: Cool, can you mention the None->global thing in the docstring?

registrar/apps/core/utils.py
```diff
@@ -27,29 +26,20 @@ def get_user_organizations(user):
...
-            match = re.match(r'\w+\.(\w+)', db_perm)
-            if match:  # pragma: no branch
-                db_perm = match.groups()[0]
-                api_permission_map[db_perm] = api_permission
-
     for db_perm in user_object_permissions + user_global_permissions:
-        if db_perm in api_permission_map:  # pragma: no branch
-            user_api_permissions.add(api_permission_map.get(db_perm))
+        if db_perm in DB_TO_API_PERMISSION_MAPPING:  # pragma: no branch
+            user_api_permissions.add(DB_TO_API_PERMISSION_MAPPING.get(db_perm))
```
kdmccormick: Nit: You're already checking `db_perm in DB_TO_API_PERMISSION_MAPPING`, so `DB_TO_API_PERMISSION_MAPPING[db_perm]` is enough

kdmccormick (approved):
Nice. Approved pending the two nits I left below.

I still think that `ProgramListView.get_queryset` could be completely written pretty easily now that  `get_user_api_permissions` exists, thus de-duping some of the logic around object-level/global permissions. Your discretion, though, on whether that's worth doing on this PR.

registrar/apps/core/utils.py
zacharis278: not sure I follow, I think we still want this.  If a db_permission is created that does not map to an API Permission we don't want to add 'None' to this list.

kdmccormick: Sorry, I phrased that poorly. Here's what I'm suggesting:
```python
if db_perm in DB_TO_API_PERMISSION_MAPPING:
    user_api_permissions.add(DB_TO_API_PERMISSION_MAPPING[db_perm])  # brackets instead of `get`
```

zacharis278: oh i see, 👍

zacharis278:
fixed a bug with permissions assigned through a django (not guardian) group.  Figured I might as well simplify that queryset function as well

registrar/apps/api/v1/views.py
```diff
@@ -95,28 +99,18 @@ def get_queryset(self):
...
-                    self.permission_filter.check(user, program))
+        if self.permission_filter:
+            required_permission = self.permission_filter
+        else:
+            required_permission = APIReadMetadataPermission
+
+        programs = (
+            program for program in programs
+            if required_permission in get_user_api_permissions(user, program).union(
+                get_user_api_permissions(user, program.managing_organization)
```
kdmccormick, line 111: Nice :+1:

PR #234, opened 2020-01-14
add pending user group model

how015:
https://openedx.atlassian.net/browse/MST-88


kdmccormick (changes requested):
I imagine that we want to only end up with one model, right? Otherwise, we may as well just have a `PendingUserProgramGroup`.

Do you have a plan for removing `PendingUserOrganizationGroup`?

Alternatively, is there a way to just rename  `PendingUserOrganizationGroup` to `PendingUserGroup`?

how015:
> I imagine that we want to only end up with one model, right? Otherwise, we may as well just have a `PendingUserProgramGroup`.
> 
> Do you have a plan for removing `PendingUserOrganizationGroup`?
> 
> Alternatively, is there a way to just rename `PendingUserOrganizationGroup` to `PendingUserGroup`?

In the UI, Pending User Organization Group is removed and replaced with Pending User Group. 

We keep the model of PendingUserOrganizationGroup just for backward compatibility, i.e. for the existing PendingUserOrganizationGroup instances, we still want to consume (and then delete) them upon user login.

We will completely remove the model of PendingUserOrganizationGroup once we are sure all PendingUserOrganizationGroup instances are deleted.

registrar/apps/core/signals.py
```diff
@@ -3,18 +3,36 @@
...
 
 
 def handle_user_post_save(sender, **kwargs):  # pylint: disable=unused-argument
     """
-    If a pending user organization group record exists, add the user to the
-    OrganizationGroup and delete the pending record.
+    If a pending user group record exists, add the user to the
+    OrganizationGroup or ProgramOrganizationGroup and delete the pending record.
+
+    For backward compatibility, if an old pending user organization group record exists,
```
zacharis278: maybe explain a bit what backwards compatibility means here.  e.g. we have existing production records yet to be consumed, but no new records will be created.  May also be worth updating the docstring on PendingOrganizationGroup as well to point out the planned deprecation.

zacharis278 (approved):

registrar/apps/core/models.py
```diff
@@ -267,6 +267,50 @@ def __repr__(self):
...
+            )
+
+        if isinstance(self.group, ProgramOrganizationGroup):  # pragma: no branch
+            return "<PendingUserGroup {ID}>: {organization_name} - {program_name} - {user_email} - {role}".format(
+                ID=self.id,
+                organization_name=self.group.granting_organization.name,
+                program_name=self.group.program.key,
+                user_email=self.user_email,
+                role=self.group.role,
+            )
```
kdmccormick, line 308: Think about what would happen in this method if we assigned `group` to a generic `Group`.

registrar/apps/core/signals.py
```diff
@@ -3,18 +3,36 @@
...
+    add the user to the OrganizationGroup and delete the pending record.
     """
     user_instance = kwargs.get("instance")
+
+    pending_groups = PendingUserGroup.objects.filter(user_email=user_instance.email)
+    for pending_group in pending_groups:
+        logger.info(
+            'add group {} into user group for user {}'.format(
```
kdmccormick: let's clear up this phrasing: `'add user {} to group {}'`

registrar/apps/core/tests/factories.py
```diff
@@ -129,3 +122,19 @@ class Meta(object):
     program = factory.SubFactory(ProgramFactory)
     granting_organization = factory.SubFactory(OrganizationFactory)
     role = ProgramReadMetadataRole.name
+
+
+class PendingUserOrganizationGroupFactory(factory.DjangoModelFactory):
+    class Meta(object):
+        model = PendingUserOrganizationGroup
```
kdmccormick, line 129: why not switch this over to `PendingUserGroup` too?

kdmccormick (changes requested):
Thanks for thinking of the migration process. It sounds good to me.

I requested some more changes, but they are fairly minor, and this should be good to go soon.

registrar/apps/core/tests/test_models.py
```diff
@@ -274,3 +275,42 @@ def test_string(self):
...
+            user_email=user_email,
+            group=self.organization_group,
+        )
+        pending_user_group_string = str(pending_user_group)
+        self.assertIn('PendingUserGroup', pending_user_group_string)
+        self.assertIn(user_email, pending_user_group_string)
+        self.assertIn(self.organization.name, pending_user_group_string)
+        self.assertIn(self.organization_group.role, pending_user_group_string)
+
+    def test_pending_program_group_string(self):
```
schenedx, line 307: Please add an unit test case for a pending user group that links to a group that do not link to either Program nor Organization. Basically, test the case when the Group is standing alone.

schenedx (commented):
Basically tagging onto what Kyle provided as feedbacks. This is close. Thank you for the refactoring

registrar/apps/core/models.py
```diff
@@ -307,6 +307,12 @@ def __str__(self):
                 role=self.group.role,
             )
 
+        return "<PendingUserGroup {ID}>: {group_name} - {user_email}".format(
+            ID=self.id,
+            group_name=self.group.name,
+            user_email=self.user_email,
+        )
```
kdmccormick, line 314: I'd include `role` too, just like the others do.

registrar/apps/core/tests/factories.py
how015: This is there to test backward compatibility.

PR #236, opened 2020-01-22
The UUID of the program on the S3 path have no dashes

schenedx:
This change make sure we point at the correct S3 path
@edx/masters-devs Please help review


michaelroytman (commented):
The change looks good, but you'll have to update the tests to handle this new file prefix. I'll approve after those tests are passing.

Thank you for fixing this!

zacharis278 (approved):

PR #237, opened 2020-01-23
Add logging to figure out why we cannot list

schenedx:
## Description

This PR is to add logging to check why we cannot retrieve reports list from the S3 buckets.
This helps debug the bug where the reports API returns empty reports.


schenedx:
It seems like this PR is not needed. Closing for now.

PR #243, opened 2020-01-29
Remove the code associated with PendingUserOrganizationGroup model

schenedx:
## Description
Remove the no longer used PendingUserOrganizationGroup model. All the data that was in this table in production, are now already moved to the new PendingUserGroup model.
This PR do not include any migration. So this is 1 out of 2 PRs I will create. This one will be deployed first.
@edx/masters-devs Please help review

### DIFF
--- a/registrar/apps/core/models.py
+++ b/registrar/apps/core/models.py
@@ -233,43 +233,6 @@ class ProgramOrganizationGroup(Group):
         return 'ProgramOrganizationGroup: {} role={}'.format(self.program.title, self.role)  # pragma: no cover
 
 
-class PendingUserOrganizationGroup(TimeStampedModel):
-    """
-    PendingUserOrganizationGroup is now deprecated, and is replaced by PendingUserGroup
-    which supports both Organization Group and Program Group.
-
-    Organization Membership model for user who have not yet been created in the user table
-
-    .. pii:: stores the user email address field for pending users.
-             The pii data gets deleted after a real user gets created
-    .. pii_types:: email_address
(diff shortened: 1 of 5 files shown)


registrar/apps/core/tests/test_signals.py
```diff
@@ -110,11 +105,5 @@ def test_no_pending_user_organization_group(self):
         )
         self.assertEqual(len(user.groups.all()), 0)
 
-    def _assert_deletion(self):
```
kdmccormick, line 113: Why delete these assertions?

If it's an implementation issue, you could pass in `model_class` as an argument to this, and then do
```python
    with self.assertRaises(model_class.DoesNotExist):
        ....
```

kdmccormick (approved):
Nice cleanup.

Approved pending my comment and codecov, which is being failed by `OrganizationGroup.__str__` being uncovered.

registrar/apps/core/tests/test_signals.py
schenedx: Good catch. Will revert

how015 (approved):

how015:
LGTM

registrar/apps/core/models.py
```diff
@@ -233,43 +233,6 @@ def __str__(self):
         return 'ProgramOrganizationGroup: {} role={}'.format(self.program.title, self.role)  # pragma: no cover
 
 
-class PendingUserOrganizationGroup(TimeStampedModel):
```
iloveagent57, line 236: Goodbye, old friend...

PR #246, opened 2020-01-31
Merge common+core apps; use relative imports

kdmccormick:
@edx/masters-devs 
https://openedx.atlassian.net/browse/MST-76

### DIFF
--- a/registrar/apps/api/exceptions.py
+++ b/registrar/apps/api/exceptions.py
@@ -4,10 +4,7 @@ Custom exceptions extending those definined by DRF.
 from rest_framework import status
 from rest_framework.exceptions import APIException
 
-from registrar.apps.api.constants import (
-    ENROLLMENT_WRITE_MAX_SIZE,
-    UPLOAD_FILE_MAX_SIZE,
-)
+from .constants import ENROLLMENT_WRITE_MAX_SIZE, UPLOAD_FILE_MAX_SIZE
 
 
 class EnrollmentPayloadTooLarge(APIException):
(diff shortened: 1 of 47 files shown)


schenedx (commented):
Why are you moving all the import statements from absolute path to relative path? What benefits do we gain from doing relative path? Is it a standard written somewhere?

kdmccormick:
@schenedx  Fair question. The [edX style guide](https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html) is silent on the topic. Here is what [PEP-8](https://www.python.org/dev/peps/pep-0008/) says:

----
Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path):
```python
import mypkg.sibling
from mypkg import sibling
from mypkg.sibling import example
```
However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose:
```python
from . import sibling
from .sibling import example
```
Standard library code should avoid complex package layouts and always use absolute imports.

----

I feel like for same-app imports, an absolute import is overly verbose. And, more importantly, I like that `isort` sorts absolute imports separate from relative imports, making it more obvious what you're importing from the *same app* versus *some other app*. My hope is that this might make us more cognizant of the dependencies we are creating between apps.

schenedx (approved):
Understand. I feel like this style change might not be followed or understood by other engineers. Can we make sure this is communicated in a very visible place?
Thank you for taking the time to update all the imports.

kdmccormick:
@schenedx Good point. Going to merge this now to avoid nasty merge conflicts, but I have a work-in-progress document including what you mentioned that I'll post as a PR shortly.

PR #252, opened 2020-02-06
Add logging to Filestore.listdir

kdmccormick:
@edx/masters-devs 
https://openedx.atlassian.net/browse/MST-133

List reports on Stage is giving 500s; this will help me figure out why.


schenedx (approved):
👍

PR #253, opened 2020-02-06
Refactor Disco cache to tolerate inter-service errors 

kdmccormick:
@edx/masters-devs 

The goals of this PR are to:
* Make sure Registrar still has base-line functionality when Discovery is unavailable or is missing a program.
* Simplify the logic around Discovery caching.

Finally, here is a follow-up PR that does some fairly simple (but diff-heavy, hence the separation) reorganization: https://github.com/edx/registrar/pull/256

Ticket: https://openedx.atlassian.net/browse/MST-131

### DIFF
--- a/registrar/apps/api/v1/mixins.py
+++ b/registrar/apps/api/v1/mixins.py
@@ -21,9 +21,9 @@ from rest_framework.status import (
 )
 
 from registrar.apps.core import permissions as perms
+from registrar.apps.core.data import DiscoveryProgram
 from registrar.apps.core.filestore import get_enrollment_uploads_filestore
 from registrar.apps.core.jobs import start_job
-from registrar.apps.core.models import Program
 from registrar.apps.enrollments.data import (
     write_course_run_enrollments,
     write_program_enrollments,
@@ -141,11 +141,14 @@ class ProgramSpecificViewMixin(AuthMixin):
     def program(self):
         """
         The program specified by the `program_key` URL parameter.
+
+        Loads a DiscoveryProgram, which gives access to fields from Discovery
+        such as title, program type, etc.
         """
         program_key = self.kwargs['program_key']
         try:
-            return Program.objects.get(key=program_key)
-        except Program.DoesNotExist:
+            return DiscoveryProgram.objects.get(key=program_key)
+        except DiscoveryProgram.DoesNotExist:
             self.add_tracking_data(failure='program_not_found')
             raise Http404()
 
@@ -174,7 +177,7 @@ class CourseSpecificViewMixin(ProgramSpecificViewMixin):
         parameter is not part of self.program.
         """
         provided_course_id = self.kwargs['course_id']
-        real_course_run = self.program.discovery_program.find_course_run(
+        real_course_run = self.program.find_course_run(
             provided_course_id
         )
         if not real_course_run:
--- a/registrar/apps/core/data.py
+++ b/registrar/apps/core/data.py
@@ -3,17 +3,16 @@ Module for syncing data with external services.
 """
 import logging
 from collections import namedtuple
-from datetime import datetime
 from posixpath import join as urljoin
+from uuid import UUID
 
 from django.conf import settings
 from django.core.cache import cache
-from django.http import Http404
 from edx_rest_api_client import client as rest_client
 from requests.exceptions import HTTPError
-from rest_framework.status import HTTP_404_NOT_FOUND
 
 from .constants import PROGRAM_CACHE_KEY_TPL, PROGRAM_CACHE_TIMEOUT
+from .models import Program
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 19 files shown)


registrar/apps/api/v1/mixins.py
```diff
@@ -141,11 +141,14 @@ class ProgramSpecificViewMixin(AuthMixin):
...
+        Loads a DiscoveryProgram, which gives access to fields from Discovery
+        such as title, program type, etc.
         """
         program_key = self.kwargs['program_key']
         try:
-            return Program.objects.get(key=program_key)
-        except Program.DoesNotExist:
+            return DiscoveryProgram.objects.get(key=program_key)
+        except DiscoveryProgram.DoesNotExist:
             self.add_tracking_data(failure='program_not_found')
```
Andytr1, line 152: tiny nit: you changed everything to discoveryprogram. Would it make sense to log 'discoveryprogram_not_found' instead?

registrar/apps/core/data.py
```diff
@@ -25,154 +24,234 @@
...
+
+        Overrides `Model.from_db`.
+        """
+        instance = super().from_db(db, field_names, values)
+        cls.get_program_data(instance.discovery_uuid)  # pylint: disable=no-member
+        return instance
 
     @classmethod
-    def get(cls, program_uuid, client=None):
+    def get_program_data(cls, program_uuid):
```
Andytr1, line 61: how about get_program_data_json ? since we are getting json representation to make it more explicit to consumers?

registrar/apps/core/data.py
```diff
@@ -25,154 +24,234 @@
...
+        if not isinstance(program_data, dict):
+            program_data = cls._fetch_discovery_program_data(program_uuid)
+            cache_value = program_data if isinstance(program_data, dict) else {}
+            cache.set(key, cache_value, PROGRAM_CACHE_TIMEOUT)
+        return program_data
 
-    @classmethod
-    def read_from_discovery(cls, program_uuid, client=None):
+    @staticmethod
+    def _fetch_discovery_program_data(program_uuid):
```
Andytr1, line 98: same nit as above. To me 'data' means data/information, but we are returning json representation here. Perhaps add json somewhere in the method's name?

registrar/apps/core/data.py
```diff
@@ -25,154 +24,234 @@
...
+        return self.discovery_data.get('type')
+
+    @property
+    def is_enrollment_enabled(self):
+        """
+        Return whether enrollment is enabled for this program.
+
+        Falls back to False if required data unavailable.
+        """
+        return self.program_type == 'Masters'
```
Andytr1, line 168: Is this correct? It looks like the check is if the program_type is Masters.

registrar/apps/core/tasks.py
```diff
@@ -36,7 +36,7 @@ def _get_program(job_id, program_key):
     Load a Program by key. Fails job and returns None if key invalid.
     """
     try:
-        return Program.objects.get(key=program_key)
-    except Program.DoesNotExist:
+        return DiscoveryProgram.objects.get(key=program_key)
+    except DiscoveryProgram.DoesNotExist:
         post_job_failure(job_id, "Bad program key: {}".format(program_key))
```
Andytr1, line 41: Would it make sense to log "Bad discovery program key..." instead?

registrar/apps/core/data.py
kdmccormick: Yup, that's correct. I'll add a comment that clarifies that Registrar enrollment is enabled for only and all Master's programs.

registrar/apps/core/data.py
kdmccormick: Not a bad idea. I'll consider this.

registrar/apps/api/v1/mixins.py
kdmccormick: I'd rather not alter our event schema now.

Also, Program vs. DiscoveryProgram is a Registrar implementation detail - `discoveryprogram` would mean nothing to someone doing analysis.

Andytr1 (commented):
Looks good. Just a few nits with one question.

registrar/apps/core/tasks.py
kdmccormick: DiscoveryProgram is a proxy around Program, which is the actual entity. If this `get` fails, then it's because the Program doesn't exist in the database.

registrar/apps/core/data.py
kdmccormick: Will consider; thanks for the feedback.

registrar/apps/core/data.py
kdmccormick: Fixing this forward in https://github.com/edx/registrar/pull/256 to avoid dealing with merge conflicts (see commit "Clarify is_enrollment_enabled logic")

registrar/apps/core/data.py
kdmccormick: I agree that `_data` is really vague and could refer to pretty much any information. However, I shied away from using `_json`, as it focuses more on the format of the data and less on what the data actually represents.

I renamed this function to `_fetch_discovery_program_details`, and related functions/variables likewise. I think that "details" communicates that we are getting *extra* data from Discovery, as opposed to the essential data, which is in the DB.

I also added bits to the docstrings that clarifies that it's returning a JSON-like dict.

The changes are in the follow-up PR: https://github.com/edx/registrar/pull/256

registrar/apps/core/data.py
kdmccormick: see https://github.com/edx/registrar/pull/253#discussion_r377653007

schenedx (approved):
This is a lot of work! Good job.

PR #255, opened 2020-02-10
Make repr() and str() of models consistent

kdmccormick:
I needed to change a couple our models' `__str__` and `__repr__` definitions as pre-work for [MST-131](https://openedx.atlassian.net/browse/MST-131). I decided to go ahead and make them all consistent.

The new test cases are looser than than old ones. I think that's OK, because as long as the `str()`s and `repr()`s of our model instances contain certain essential information and don't raise exceptions, then there shouldn't be a problem.

Next PR: https://github.com/edx/registrar/pull/253

@edx/masters-devs

### DIFF
--- a/registrar/apps/core/models.py
+++ b/registrar/apps/core/models.py
@@ -2,7 +2,6 @@
 
 from django.contrib.auth.models import AbstractUser, Group
 from django.db import models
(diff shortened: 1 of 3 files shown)


schenedx (approved):
👍

PR #257, opened 2020-02-11
Omit enrollment permissions for enrollment-disable programs

kdmccormick:
work-in-progress, on hold for now

Made while investigating https://openedx.atlassian.net/browse/MST-146


kdmccormick:
Closed because Michael is handling this in [MST-198](https://openedx.atlassian.net/browse/MST-198).

PR #261, opened 2020-02-12
Adding dockerfile

coryleeio:


kdmccormick:
I love how simple this is.

fredsmith (approved):

Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM ubuntu:xenial
+
+RUN apt update && \
+  apt install -qy git-core language-pack-en libmysqlclient-dev libssl-dev python3.5 python3-pip python3.5-dev && \
+  pip3 install --upgrade pip setuptools && \
+  rm -rf /var/lib/apt/lists/*
```
kdmccormick: `apt` warns that its interface isn't stable. Probably better to use `apt-get` here.
```suggestion
RUN apt-get update && \
  apt-get install -qy git-core language-pack-en libmysqlclient-dev libssl-dev python3.5 python3-pip python3.5-dev && \
  pip3 install --upgrade pip setuptools && \
  rm -rf /var/lib/apt/lists/*
```

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+RUN ln -s /usr/bin/pip3 /usr/bin/pip
+RUN ln -s /usr/bin/python3 /usr/bin/python
+
+RUN locale-gen en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US:en
+ENV LC_ALL en_US.UTF-8
+
+# Create python venv
```
kdmccormick: Doesn't look like we create a virtual env.

Not that we need to... I think installing deps right into the image should be fine.
```suggestion
```

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN mkdir -p /edx/app/registrar
+
+WORKDIR /edx/app/registrar
+COPY requirements /edx/app/registrar/requirements
+RUN pip install -r requirements/production.txt
+
+RUN useradd -m --shell /bin/false app
+USER app
+
+EXPOSE 8734
```
kdmccormick, line 25:
```suggestion
EXPOSE 18734
```
Keeping the port as `18734` would prevent needing to change devstack. Do you think we should switch to 8734?

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+CMD gunicorn -c /edx/app/registrar/registrar/docker_gunicorn_configuration.py --bind=0.0.0.0:8734 --workers 2 --max-requests=1000 registrar.wsgi:application
+
+COPY . /edx/app/registrar
```
kdmccormick, line 28: Consider creating `.dockerignore` as a soft-link to `.gitignore`. I built this locally and then very nearly pushed my Segment key to Dockerhub.

Plus, making the build context smaller will make this  build a bit quicker in people's local checkouts.

registrar/docker_gunicorn_configuration.py
```diff
@@ -0,0 +1,8 @@
+timeout = 300
```
kdmccormick, line 1: pylint wants a docstring

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US:en
+ENV LC_ALL en_US.UTF-8
+
+# Create python venv
+RUN mkdir -p /edx/app/registrar
+
+WORKDIR /edx/app/registrar
+COPY requirements /edx/app/registrar/requirements
+RUN pip install -r requirements/production.txt
```
kdmccormick, line 20:
```suggestion
RUN make production-requirements
```

kdmccormick (commented):
Nice. I'd love to get to a point where we can make this the image Registrar uses for devstack, CI, and even deployment.

When I build and run the container:
```bash
$ docker build -t kdmccormick96/registrar:latest .
$ docker run --name reg kdmccormick96/registrar:latest
[2020-04-03 20:48:39 +0000] [6] [INFO] Starting gunicorn 20.0.4
[2020-04-03 20:48:39 +0000] [6] [INFO] Listening at: http://0.0.0.0:8734 (6)
[2020-04-03 20:48:39 +0000] [6] [INFO] Using worker: gevent
[2020-04-03 20:48:39 +0000] [9] [INFO] Booting worker with pid: 9
[2020-04-03 20:48:39 +0000] [10] [INFO] Booting worker with pid: 10
```
and then hit localhost:8734 in a browser, I see "Unable to connect".

PR #266, opened 2020-02-18
Integrated tox to support testing multiple versions of Django -- BOM-1284

mraarif:
## Description
Relevant JIRA issue can be found [here](https://openedx.atlassian.net/browse/BOM-1284).


tox.ini
```diff
@@ -0,0 +1,18 @@
+[tox]
+envlist = python3.6-django{20,21,22},quality
```
jmbowman: Should include Django 1.11 here also.

mraarif: ahh, yes, Thanks for the catch

jmbowman (approved):

awais786 (approved):

PR #268, opened 2020-02-21
BOM-1285 : Updgrade edx-lint in registrar

mzulqarnain1:
Relevant JIRA Link : [https://openedx.atlassian.net/browse/BOM-1285](https://openedx.atlassian.net/browse/BOM-1285)

### DIFF
--- a/registrar/apps/api/serializers.py
+++ b/registrar/apps/api/serializers.py
@@ -36,6 +36,9 @@ class DiscoveryProgramSerializer(serializers.ModelSerializer):
         )
 
     def get_permissions(self, program):
+        """
+        Method to get permissions of a user
+        """
         request = self.context.get('request')
         if request and hasattr(request, 'user'):
             user = request.user
--- a/registrar/apps/api/serializers.py
+++ b/registrar/apps/api/serializers.py
@@ -37,7 +37,7 @@ class DiscoveryProgramSerializer(serializers.ModelSerializer):
 
     def get_permissions(self, program):
         """
-        Method to get permissions of a user
+        Get list of permissions granted to user making the request from the context
         """
         request = self.context.get('request')
         if request and hasattr(request, 'user'):
--- a/registrar/apps/api/utils.py
+++ b/registrar/apps/api/utils.py
@@ -47,4 +47,4 @@ def to_absolute_api_url(path, *more_paths):
     path_parts = [settings.API_ROOT, path[len(PREFIX):]] + list(more_paths)
     stripped_path_parts = [part.strip('/') for part in path_parts]
     result = '/'.join(part for part in stripped_path_parts if part)
-    return result + ('/' if (path_parts[-1]).endswith('/') else '')
+    return result + ('/' if str(path_parts[-1]).endswith('/') else '')
--- a/registrar/apps/api/utils.py
+++ b/registrar/apps/api/utils.py
@@ -47,4 +47,4 @@ def to_absolute_api_url(path, *more_paths):
     path_parts = [settings.API_ROOT, path[len(PREFIX):]] + list(more_paths)
     stripped_path_parts = [part.strip('/') for part in path_parts]
     result = '/'.join(part for part in stripped_path_parts if part)
-    return result + ('/' if path_parts[-1].endswith('/') else '')
+    return result + ('/' if (path_parts[-1]).endswith('/') else '')
--- a/registrar/apps/api/v1/mixins.py
+++ b/registrar/apps/api/v1/mixins.py
@@ -70,7 +70,7 @@ class AuthMixin(TrackViewMixin):
         if isinstance(self.permission_required, str):  # pragma: no branch
             return [self.permission_required]  # pragma: no cover
         elif isinstance(self.permission_required, Iterable):
-            return [p for p in self.permission_required]
+            return self.permission_required
         else:  # pragma: no cover
             raise ImproperlyConfigured(
                 'permission_required must be a string or iterable; ' +
@@ -115,6 +115,7 @@ class AuthMixin(TrackViewMixin):
             self.add_tracking_data(missing_permissions=list(missing_permissions))
             self._unauthorized_response()
 
+    # pylint: disable=missing-docstring
     @staticmethod
     def _has_permission_on_any(user, perm, objects):
         for obj in objects:
@@ -128,8 +129,8 @@ class AuthMixin(TrackViewMixin):
         """
         if self.raise_404_if_unauthorized:
             raise Http404()
-        else:
-            raise PermissionDenied()
+
+        raise PermissionDenied()
 
 
 class ProgramSpecificViewMixin(AuthMixin):
@@ -194,7 +195,7 @@ class CourseSpecificViewMixin(ProgramSpecificViewMixin):
         return self.course_run.external_key
 
 
-class JobInvokerMixin(object):
+class JobInvokerMixin:
     """
     A mixin for views that invoke jobs and return ID and status URL.
     """
@@ -255,8 +256,9 @@ class EnrollmentMixin(ProgramSpecificViewMixin):
             # Raise exception if the program (MM at the moment) is not
             # available for enrollments related API endpoints
             raise PermissionDenied(
-                'Cannot access enrollment endpoints with program [%s] whose enrollments are disabled',
-                self.program.key
+                'Cannot access enrollment endpoints with program [{}] whose enrollments are disabled'.format(
+                    self.program.key
+                )
             )
 
         super().check_permissions(request)
--- a/registrar/apps/api/v1/mixins.py
+++ b/registrar/apps/api/v1/mixins.py
@@ -115,9 +115,9 @@ class AuthMixin(TrackViewMixin):
             self.add_tracking_data(missing_permissions=list(missing_permissions))
             self._unauthorized_response()
 
-    # pylint: disable=missing-function-docstring
     @staticmethod
     def _has_permission_on_any(user, perm, objects):
+        # pylint: disable=missing-function-docstring
         for obj in objects:
             if perm.check(user, obj):
                 return True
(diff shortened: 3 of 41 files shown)


registrar/apps/api/serializers.py
```diff
@@ -36,6 +36,9 @@ class Meta:
         )
 
     def get_permissions(self, program):
+        """
+        Method to get permissions of a user
```
kdmccormick: "Get list of permissions granted to user making the request from the context"

registrar/apps/api/utils.py
```diff
@@ -47,4 +47,4 @@ def to_absolute_api_url(path, *more_paths):
     path_parts = [settings.API_ROOT, path[len(PREFIX):]] + list(more_paths)
     stripped_path_parts = [part.strip('/') for part in path_parts]
     result = '/'.join(part for part in stripped_path_parts if part)
-    return result + ('/' if path_parts[-1].endswith('/') else '')
+    return result + ('/' if (path_parts[-1]).endswith('/') else '')
```
kdmccormick: I'm surprised this change is necessary

registrar/apps/api/v1/views.py
```diff
@@ -416,6 +416,7 @@ def get_queryset(self):
 
 
 class EnrollmentUploadView(JobInvokerMixin, APIView):
+    # pylint: disable=no-member
```
kdmccormick: can you put this on the specific line(s) that are causing the warning?

registrar/apps/api/v2/urls.py
```diff
@@ -5,4 +5,4 @@
 
 app_name = 'v2'
 
-urlpatterns = [pattern for pattern in v1_urlpatterns]
+urlpatterns = v1_urlpatterns
```
kdmccormick: this is a purposeful list copy - can you make it `v1_urlpatterns.copy()`?

registrar/apps/core/jobs.py
```diff
@@ -146,10 +146,9 @@ def _get_result(task_status):
         if artifacts.count() > 1:  # pragma: no cover
             logger.error(
                 'Multiple UserTaskArtifacts for job ' +
-                '(job_id = {}, UserTaskStatus.uuid = {}). '.format(
-                    task_status.task_id, task_status.uuid
-                ) +
-                'First artifact will be returned'
+                '(job_id = {}, UserTaskStatus.uuid = {}). First artifact will be returned',
+                task_status.task_id,
+                task_status.uuid
```
kdmccormick, line 151: are you sure this works? you might need to replace the `{}`s with `%s`

registrar/apps/enrollments/tasks.py
```diff
@@ -50,7 +50,7 @@ def generate_name(cls, arguments_dict):
         )
 
 
-# pylint: disable=unused-argument
+# pylint: disable=unused-argument, inconsistent-return-statements
```
kdmccormick: instead of disabling `inconsistent-return-statements`, you can fix this by changing `return None` to `return`. you can do this for all four occurrences of this warning.

registrar/apps/core/models.py
```diff
@@ -15,6 +15,7 @@
 
 
 class User(AbstractUser):
+    # pylint: disable:no-member
```
kdmccormick: can you put this on the specific line(s) it occurs on? that way, if we cause real `no-member` warnings in this class, they won't get silenced.

registrar/apps/enrollments/tests/test_tasks.py
```diff
@@ -40,6 +40,7 @@
 
 @ddt.ddt
 class ListEnrollmentTaskTestMixin(BaseTaskTestMixin):
+    # pylint: disable=assignment-from-no-return
```
kdmccormick: same comment from before - can you put this on the exact line(s)?

kdmccormick (commented):
Big thanks for these changes @mzulqarnain1 ; it's something we've been meaning to do for a while.

I have some comments below. Most are nits, but the one about the log formatting (`{}` vs `%s`) is significant.

registrar/apps/core/tests/mixins.py
```diff
@@ -57,13 +56,13 @@ def assert_failed(self, sub_message, job_id=None):
...
 
-    def test_program_not_found(self):
+    def test_program_not_found(self):  # pylint: disable=assignment-from-no-return
         task = self.spawn_task(program_key="program-nonexistant")
         task.wait()
         self.assert_failed("Bad program key")
 
 
-class S3MockEnvVarsMixin(object):
+class S3MockEnvVarsMixin:  # pylint: disable=missing-class-docstring
```
nedbat: Wouldn't it be better to write a docstring?  There's a comment below that seems interesting.

registrar/apps/core/jobs.py
mzulqarnain1: you are right. I have changed it to use %s.

registrar/apps/core/tests/mixins.py
mzulqarnain1: Added a docstring.

mzulqarnain1:
> Big thanks for these changes @mzulqarnain1 ; it's something we've been meaning to do for a while.
> 
> I have some comments below. Most are nits, but the one about the log formatting (`{}` vs `%s`) is significant.

Thanks for the review @kdmccormick and @nedbat. I have implemented the feedback and made some more changes. Can you guys have a look again. Thanks.

registrar/apps/enrollments/tasks.py
```diff
@@ -86,6 +86,7 @@ def list_program_enrollments(self, job_id, user_id, file_format, program_key):
     post_job_success(job_id, serialized, file_format)
 
 
+# pylint: disable=inconsistent-return-statements
```
kdmccormick: Similar to above, you can avoid this pylint-disable by just changing `return None` to `return`. This is because Pylint thinks the returns are inconsistent because some paths explicitly return a value (`None`) whereas others don't, even though `None` is the default return value anyway.

registrar/apps/enrollments/tasks.py
```diff
@@ -132,7 +133,7 @@ def list_course_run_enrollments(
         raise ValueError('Invalid file_format: {}'.format(file_format))
     post_job_success(job_id, serialized, file_format)
 
-
+# pylint: disable=inconsistent-return-statements
```
kdmccormick: Same comment above (`return None` -> `return`)

registrar/apps/core/tests/mixins.py
```diff
@@ -58,12 +57,15 @@ def assert_failed(self, sub_message, job_id=None):
         self.assertIn(sub_message, error_artifact.text)
 
     def test_program_not_found(self):
-        task = self.spawn_task(program_key="program-nonexistant")
+        task = self.spawn_task(program_key="program-nonexistant")  # pylint: disable=assignment-from-no-return
```
kdmccormick, line 60: an alternative approach, which avoids the pylint-disable, is to change the body of `spawn_task` above to `return 'override-me'  # pragma: no cover`

registrar/apps/enrollments/tests/test_tasks.py
```diff
@@ -70,15 +70,15 @@ def test_http_error(self, status_code):
         with patch.object(tasks.lms, self.mocked_get_enrollments_method) as mock_get_enrollments:
             error = HTTPError(request=FakeRequest('registrar.edx.org'), response=FakeResponse(status_code))
             mock_get_enrollments.side_effect = error
-            task = self.spawn_task()
+            task = self.spawn_task()  # pylint: disable=assignment-from-no-return
```
kdmccormick, line 73: making the change to `def spawn_task` that I suggested in `test/mixins.py` will allow you to remove these pylint-disables

registrar/apps/api/v1/mixins.py
```diff
@@ -115,6 +115,7 @@ def check_permissions(self, request):
             self.add_tracking_data(missing_permissions=list(missing_permissions))
             self._unauthorized_response()
 
+    # pylint: disable=missing-function-docstring
```
kdmccormick: Placing pylint-disable comments *outside* the function/class that they are supposed to operate upon will actually make them have an effect outside of that function/class. For example, imagine `y` and `z` are both undefined, and we want to silence that warning in just `some_func`:
```
# pylint: disable=undefined-variable
def some_func():
    return y

def another_func():
    return z
```
The above code will unintentionally also disable the warning for `return z` in `another_func`, because of the pylint-disable placement.

This code, however, will only silence the undefined variable warning in `some_func`:
```
def some_func():
    # pylint: disable=undefined-variable
    return y

def another_func():
    return z
```

So, please make sure all to put all function-level or class-level pylint-disables within the body of the function or class.

kdmccormick (approved):
Here are a few more comments. Please make sure that all new pylint-disable comments are within the functions/classes that they are affecting, as opposed to on top of them (my last comment explains this).

Otherwise, you're good to merge :+1: Thanks again!

registrar/apps/core/tests/mixins.py
mzulqarnain1: Following this approach generates 3 more pylint errors

`registrar/apps/core/tests/mixins.py:26:25: W0613: Unused argument 'program_key' (unused argument)`

`registrar/apps/core/tests/mixins.py:26:0: W0613: Unused argument 'kwargs' (unused-argument)`

`registrar/apps/core/tests/mixins.py:62:8: E1101: Instance of 'str' has no 'wait' member (no-member)`

Then we have to either suppress those too or write a  function definition of `spawn_task` which is not right thing to do here.

registrar/apps/api/v1/mixins.py
mzulqarnain1: Got it. Thanks for such an elaborative explanation. I have fixed it and moved the suppression in the body of function.

registrar/apps/core/tests/mixins.py
kdmccormick: your original solution is definitely the right one then :) thanks for trying mine out.

jmbowman (approved):

PR #277, opened 2020-03-17
Improve Registrar model admin

kdmccormick:
As Registrar's tables grow in number of rows, it'd be good to make these pages as useful and easy-to-navigate as possible.

@edx/masters-devs


schenedx (approved):
👍

PR #278, opened 2020-03-17
add feature toggle for allowing course staff role assignment

zacharis278:
### [MST-185](https://openedx.atlassian.net/browse/MST-185?atlOrigin=eyJpIjoiMjNlNzY5NjBkNTg4NDQ3YzliMDUwMDIyZmVkY2I4ZWMiLCJwIjoiaiJ9)

Opt in toggle at the organization for allowing the assignment of course staff while uploading enrollments.


michaelroytman:
I wonder if using a Django waffle flag together with a user group (e.g.a GTx one that allows read/write enrollments, in this case) would work here. It might be more light weight than adding a field to this model (i.e. we wouldn't have to remove this model field once this feature is rolled out to all partner organizations; it would be as simple as deleting the waffle flag). Is that something you've considered?

schenedx (approved):
👍

zacharis278:
closing in favor of using a waffle flag

PR #279, opened 2020-03-19
add constant to define role management waffle flag

zacharis278:
Just adds a constant to reference for future development on role management.  This flag has been added to stage and prod and is enabled for superusers only for now.  I've tested that we can enable this for an Organization or Program group.


schenedx (approved):
👍

PR #280, opened 2020-03-19
Filter out reports with unexpected filenames

kdmccormick:
@edx/masters-devs 
https://openedx.atlassian.net/browse/MST-149


schenedx (approved):
👍

PR #282, opened 2020-03-20
Install edx-api-doc-tools behind feature flag

kdmccormick:
This PR adds the new `edx-api-doc-tools`-based documentation system to Registrar behind the `ENABLE_NEW_API_DOCS` feature flag, which is set to `False` by default.

(For the reasoning on what the new API doc system is and why I'd like to switch to it, see [MST-107](https://openedx.atlassian.net/browse/MST-107))

I am choosing to roll this out using a feature flag because writing high-quality new documentation will take time, and maintaining a long-running feature flag with the new docs has been problematic so far (every time I try to get started writing, I realize that my PR has fallen out of date with Registrar).

This PR should not change stage, production, or devstack environments by default. However, the new doc tools package will be installed in `base.txt` and the new `api-docs/` views are hit through our test suite, which will make adding new documentation as we go much easier.

How to see what the new API docs look like in this PR:
1. Create a file in your `registrar/` checkout called `settings/private.py`
2. Add the line `ENABLE_NEW_API_DOCS = True`
3. Load http://localhost:18734/api-docs/ (trailing slash is required, for now)

@edx/masters-devs


registrar/apps/api/tests/test_views.py
```diff
@@ -71,3 +76,22 @@ def assert_paths(self, response, expected_paths):
...
+class APIDocViewTest(APITestCase):
+    """ Tests for accessing api-docs """
+    path = '/api-docs/'
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.user = UserFactory()
+
+    def test_docs_access_logged_out(self):
```
schenedx, line 128: Is there a way to test whether the rendered web page contains the new documentation?

registrar/apps/api/v1/views.py
```diff
@@ -68,21 +75,29 @@
...
+        **SCHEMA_COMMON_RESPONSES,
+    },
+)
 class ProgramListView(AuthMixin, TrackViewMixin, ListAPIView):
     """
     A view for listing program objects.
 
     Path: /api/[version]/programs?org={org_key}
-
-    All programs within organization specified by `org_key` are returned.
```
schenedx, line 78: If these are removed, what happens when the `ENABLE_NEW_API_DOCS` is false? Would this impact the existing `/api-docs` view?

schenedx (commented):
Is the idea that when you think the new API system is ready, you would just remove the `ENABLE_NEW_API_DOCS` setting?

I only have question in my comments right now.

registrar/apps/api/tests/test_views.py
kdmccormick: Good call. I will do that.

registrar/apps/api/v1/views.py
kdmccormick: It would not. The existing `/api-docs` view is generated entirely from the hand-written `api.yml`.

schenedx (approved):
LGTM

kdmccormick:
Closing in favor of https://github.com/edx/registrar/pull/283.

PR #283, opened 2020-03-23
Install edx-api-doc-tools alongside old API docs

kdmccormick:
(A simpler approach to https://github.com/edx/registrar/pull/282)

New automatically-generated API docs are installed alongside existing ones under route `/api-docs/new`

Existing hand-written API docs will be removed once new API docs have been fleshed out with annotationsin order to make them as informative as the existing ones.

@edx/masters-devs 
https://openedx.atlassian.net/browse/MST-107


scripts/generate_mock_files.sh
```diff
@@ -1,36 +0,0 @@
-#!/bin/sh
```
kdmccormick, line 1: Deleting this in a separate commit as unrelated clean-up.

schenedx (approved):
👍

PR #285, opened 2020-03-25
Bulk-fetch programs from Discovery for performance

kdmccormick:
*work-in-progress*

Currently, when the program details cache is cold,
we repeatedly call Course Discovery to fetch program
details.

This PR, instead, makes it so we fetch all non-cached
programs at once using the `uuids` filter on Discovery's
/api/v1/programs endpoint.

https;//openedx.atlassian.net/MST-191


kdmccormick:
Closing for the time being.

PR #286, opened 2020-03-30
Prevent mysterious celery routing by explicitly specifying default q

matthugs:
JIRA:MST-196

## Description

Registrar is ending up putting tasks in a queue it ostensibly hasn't been configured to use as its default for quite a while on edX's staging environment. This is an attempt to address that, but should smell a little like "I can't believe it's not butter" because it will be a bit unclear why it worked.


kdmccormick (approved):
:butterfly:

PR #287, opened 2020-03-31
Downgrade to avoid permission loss

matthugs:
JIRA:ARCHBOM-1078


jmbowman (approved):
I think that should do it.

PR #288, opened 2020-04-01
Exclude CSV upload API docs; constrain cryptography; make upgrade.

kdmccormick:
I imagine that we don't want the CSV upload endpoints to be part of our official Registrar API documentation. So, this PR install the new version of `edx-api-doc-tools` in order to use the `@exclude_schema_for_all` decorator on the CSV upload views. Note that this only affects the new API docs (https://registrar.edx.org/api-docs/new)

Also, on @robrap  's [suggestion](https://edx-internal.slack.com/archives/CD89W7Z7F/p1585937317276800), we constrain `cryptography` to `<2.9` for now before running `make upgrade`.

@edx/masters-devs


zacharis278 (approved):

schenedx (approved):
👍

requirements/constraints.txt
```diff
@@ -29,3 +29,7 @@ Django>=2.2,<3.0
 
 # This package is pinned due to openedx.atlassian.net/browse/ARCHBOM-1078
 social-auth-core<3.3.0
+
+# Need to review backward incompatible changes before upgrading.
+# See https://github.com/pyca/cryptography/blob/master/CHANGELOG.rst#29---2020-04-02
+cryptography<2.9
```
robrap: You missed my follow-up. We determined it is safe.  You can drop this constraint.  Thank you.

PR #290, opened 2020-04-06
Add course_staff field to write course enrollment api in registrar

how015:
## Description

MST-186.
For now, only add a successful write course enrollment unit test, waiting for comments on my handling of waffle flag mock.
Per https://waffle.readthedocs.io/en/stable/testing/automated.html, looks like we can only make the flag a deterministic value for unit test.

@edx/masters-devs

### DIFF
--- a/registrar/apps/api/v1/mixins.py
+++ b/registrar/apps/api/v1/mixins.py
@@ -4,6 +4,7 @@ Mixins for the public V1 REST API.
 import uuid
 from collections.abc import Iterable
 
+import waffle
 from django.core.exceptions import ImproperlyConfigured, PermissionDenied
 from django.http import Http404
 from django.utils.functional import cached_property
@@ -272,7 +273,7 @@ class EnrollmentMixin(ProgramSpecificViewMixin):
         Does program enrollments if `course_id` is None.
         Does course run enrollments otherwise.
         """
-        self.validate_enrollment_data(self.request.data)
+        self.validate_enrollment_data(self.request, course_id)
         if course_id:
             good, bad, results = write_course_run_enrollments(
                 self.request.method,
@@ -295,10 +296,11 @@ class EnrollmentMixin(ProgramSpecificViewMixin):
             self.add_tracking_data(failure='unprocessable_entity')
         return Response(results, status=status)
 
-    def validate_enrollment_data(self, enrollments):
+    def validate_enrollment_data(self, request, course_id=None):
         """
         Validate enrollments request body
         """
+        enrollments = request.data
         if not isinstance(enrollments, list):
             self.add_tracking_data(failure='bad_request')
             raise ValidationError('expected request body type: List')
@@ -323,3 +325,14 @@ class EnrollmentMixin(ProgramSpecificViewMixin):
                 raise ValidationError(
                     'expected request dicts to have string value for "status"'
                 )
+            if enrollment.get('course_staff') is not None:
+                if not isinstance(enrollment.get('course_staff'), bool):
+                    self.add_tracking_data(failure='bad_request')
+                    raise ValidationError(
+                        'expected request dicts to have boolean value for "course_staff"'
+                    )
+                if course_id:
+                    if not waffle.flag_is_active(request, 'enable_course_role_management'):
+                        raise PermissionDenied('"course_staff" not accepted since role assignment is not enabled')
+                else:
+                    raise PermissionDenied('"course_staff" field is for course only')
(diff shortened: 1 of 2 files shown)


registrar/apps/api/v1/tests/test_views.py
```diff
@@ -1618,14 +1621,66 @@ def test_successful_program_course_enrollment_write(self, use_external_course_ke
...
+            self.student_course_enrollment('active', '002', False),
+            self.student_course_enrollment('inactive', '003', True),
+        ]
+
+        with self.assert_tracking(
+                user=self.stem_admin,
+                program_key=self.cs_program.key,
+                course_id=course_id,
+        ):
+            with override_flag('enable_course_role_management', active=True):
```
zacharis278: While this does work, have you tried to create a flag that would be tied to an Organization/Program group?  I think that would more closely reflect the way this will be configured for real.

If that seems too kludgy, I'd recommend using the override_flag decorator just to clean up needing to add a with statement to every test.

how015: There are two ways to manage waffle flag per [this page](https://waffle.readthedocs.io/en/stable/usage/cli.html#managing-waffle-data-from-the-command-line): "Aside the Django admin interface, you can use the command line tools to manage all your waffle objects."

Correct me if I am wrong: Since neither Django admin interface or command line can be used in unit test, we can't set up a mock group and then activate the waffle flag on it in unit test. So I think our only option is override_flag?

registrar/apps/api/v1/mixins.py
```diff
@@ -274,6 +275,11 @@ def handle_enrollments(self, course_id=None):
         """
         self.validate_enrollment_data(self.request.data)
         if course_id:
+            if not waffle.flag_is_active(self.request, 'enable_course_role_management'):
```
schenedx: Why not do these input validation in the function `self.validate_enrollment_data` instead? Why do it right after it? I know the waffle check only applies if the course_id is present. That can also be checked in the `self.validate_enrollment_data` function, right?

schenedx (commented):
The code change is looking good.
I think you just need to add more unit test to associate different types of permission groups to the waffle flag and ensure the expected result. I am glad the actual code change is actually minimal on the `mixins.py` file

zacharis278:
Nice, the payload for this looks good and matches what LMS will be expecting.  I'd second both of Simon's points on moving the validation and additional testing on groups.

registrar/apps/api/v1/mixins.py
how015: You are right Simon, working on it.

schenedx (changes requested):
Can you add another unit test making sure if the waffle flag is turned on for 1 program, doing course_staff on another program who did not get associated with the waffle flag would return 403?
Could you also please fix your pep8 failure? Make sure the command `make quality` passes before you push your code?

registrar/apps/api/v1/mixins.py
```diff
@@ -323,3 +325,11 @@ def validate_enrollment_data(self, enrollments):
                 raise ValidationError(
                     'expected request dicts to have string value for "status"'
                 )
+            if course_id and enrollment.get('course_staff') is not None:
+                if not isinstance(enrollment.get('course_staff'), bool):
```
kdmccormick:
```suggestion
            course_staff_flag = enrollment.get('course_staff')
            if course_id and course_staff_flag is not None:
                if not isinstance(course_staff_flag, bool):
```
Style nit: avoid local code duplication

registrar/apps/api/v1/mixins.py
```diff
@@ -295,10 +296,11 @@ def handle_enrollments(self, course_id=None):
             self.add_tracking_data(failure='unprocessable_entity')
         return Response(results, status=status)
 
-    def validate_enrollment_data(self, enrollments):
+    def validate_enrollment_data(self, request, course_id=None):
         """
         Validate enrollments request body
         """
+        enrollments = request.data
```
kdmccormick, line 303: Why do we need to pass in the whole request?

kdmccormick (commented):
Nice job. It looks like Simon and Zach have mentioned everything else that I would have other than two minor comments.

registrar/apps/api/v1/mixins.py
schenedx: This is because the waffle flag check requires the whole request. See line 334

schenedx (approved):
I think the code LGTM. Please address Kyle's NIT. Thank you!

PR #291, opened 2020-04-06
Add migration file; was missing due to Django 2.2 upgrade

kdmccormick:
The Django 2.2 resulted in a new migration file being required, but we could not create it until the Django 1.11 through Django 2.1 tests were removed.

Now that those tests have been removed in #294 , we can add the migration file.

@edx/masters-devs-cosmonauts


registrar/apps/core/migrations/0011_tweaks_for_django_22.py
```diff
@@ -0,0 +1,29 @@
...
+        migrations.CreateModel(
+            name='DiscoveryProgram',
+            fields=[
+            ],
+            options={
+                'indexes': [],
+                'constraints': [],
+                'proxy': True,
+            },
+            bases=('core.program',),
```
schenedx, line 22: This shouldn't have any database level change correct? It's interesting we need `proxy` option here.

schenedx (approved):
👍 Thank you for closing the loop here.

registrar/apps/core/migrations/0011_tweaks_for_django_22.py
kdmccormick: No database level changes.

The `proxy` option shows that DiscoveryProgram is just a proxy model to Program, as opposed to a model subclass with its own table (like Degree and Program).

PR #295, opened 2020-04-07
mroytman/mst 199 enrollments disabled permissions

michaelroytman:
## Description

[JIRA ticket](https://openedx.atlassian.net/browse/MST-198)

> The /api/v2/programs endpoint, which is called by Program Console, should return a list of programs and a list of permissions available to the user for each program.
> Because MicroMaster’s programs are considered to have “enrollment disabled”, they should not return write_enrollments and read_enrollments in their list of permissions.
> However, if someone has write/read_enrollments permission assigned through a MM Program or an Organization that runs a MM program, the endpoint will report write/read_enrollments for that MM program.

### DIFF
--- a/registrar/apps/api/serializers.py
+++ b/registrar/apps/api/serializers.py
@@ -8,7 +8,7 @@ from rest_framework import serializers
 from user_tasks.models import UserTaskStatus
 
 from registrar.apps.core.proxies import DiscoveryProgram
-from registrar.apps.core.utils import get_user_api_permissions
+from registrar.apps.core.utils import get_effective_user_program_api_permissions
 from registrar.apps.enrollments.constants import (
     COURSE_ENROLLMENT_STATUSES,
     PROGRAM_ENROLLMENT_STATUSES,
@@ -41,12 +41,8 @@ class DiscoveryProgramSerializer(serializers.ModelSerializer):
         """
         request = self.context.get('request')
         if request and hasattr(request, 'user'):
-            user = request.user
-            user_permissions = get_user_api_permissions(
-                user, program
-            ).union(
-                get_user_api_permissions(user, program.managing_organization)
-            )
+            user_permissions = get_effective_user_program_api_permissions(request.user, program)
+
             return sorted(permission.name for permission in user_permissions)
         else:
             return []
--- a/registrar/apps/api/tests/test_serializers.py
+++ b/registrar/apps/api/tests/test_serializers.py
@@ -36,21 +36,21 @@ class DiscoveryProgramSerializerTests(TestCase):
             PROGRAM_CACHE_KEY_TPL.format(uuid=self.program.discovery_uuid),
             {
                 'title': 'test-program',
-                'type': 'masters',
+                'type': 'Masters',
                 'curricula': [],
-            },
+            }
         )
 
-    @mock.patch('registrar.apps.api.serializers.get_user_api_permissions')
-    def test_get_user_permissions(self, fake_get_perms):
+    @mock.patch('registrar.apps.api.serializers.get_effective_user_program_api_permissions')
+    def test_serializer(self, fake_get_perms):
         """
-        Program permissions should be populated with the union of all
-        permissions assigned at that program and its managing organization
+        Test that the serializer should return output of get_effective_user_program_api_permissions
+        for permissions.
         """
-        fake_get_perms.side_effect = [
-            set([APIWriteEnrollmentsPermission]),  # mocked permissions on program
-            set([APIReadMetadataPermission]),  # mocked permissions on org
-        ]
+        fake_get_perms.return_value = {
+            APIWriteEnrollmentsPermission,
+            APIReadMetadataPermission,
+        }
         program = DiscoveryProgramSerializer(
             self.program,
             context={
--- a/registrar/apps/core/tests/test_utils.py
+++ b/registrar/apps/core/tests/test_utils.py
@@ -1,15 +1,20 @@
 """ Tests for Core utils.py """
 import ddt
 from django.contrib.auth.models import Group
+from django.core.cache import cache
 from django.test import TestCase
 from guardian.shortcuts import assign_perm
 from rest_framework.exceptions import ValidationError
 
+from ..constants import PROGRAM_CACHE_KEY_TPL
 from ..permissions import (
     ORGANIZATION_READ_ENROLLMENTS,
     ORGANIZATION_READ_METADATA,
     ORGANIZATION_READ_REPORTS,
     ORGANIZATION_WRITE_ENROLLMENTS,
+    PROGRAM_READ_ENROLLMENTS,
+    PROGRAM_READ_METADATA,
+    PROGRAM_WRITE_ENROLLMENTS,
     APIReadEnrollmentsPermission,
     APIReadMetadataPermission,
     APIReadReportsPermission,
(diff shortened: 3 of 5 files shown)


registrar/apps/core/utils.py
```diff
@@ -26,12 +28,47 @@ def get_user_organizations(user):
...
 def get_user_api_permissions(user, obj=None):
     """
     Returns a set of all APIPermissions granted to the user on a
     provided object instance. This includes permissions granted though a
     global permission or role. If no object is passed only global permissions
     will be returned.
+
+    You should not use this function directly if you want a set of a user's effective
+    APIPermissions, as this function does not account for programs for which enrollments
+    are disabled.
```
schenedx: #NIT in the comment, make sense to reference the new `get_effective_user_program_api_permissions` function

registrar/apps/core/tests/test_utils.py
```diff
@@ -122,6 +129,91 @@ def test_get_api_permissions(self):
...
+        assign_perm(PROGRAM_WRITE_ENROLLMENTS, cls.user2, cls.masters_program)
+
+        assign_perm(PROGRAM_READ_METADATA, cls.user2, cls.non_masters_program)
+        assign_perm(PROGRAM_READ_ENROLLMENTS, cls.user2, cls.non_masters_program)
+        assign_perm(PROGRAM_READ_ENROLLMENTS, cls.user2, cls.non_masters_program)
+
+    def test_get_effective_user_program_api_permissions(self):
+        # check effective permissions of a user on a program via organization
+        # and global permissions
+        perms = get_effective_user_program_api_permissions(self.user1, self.masters_program)
```
schenedx, line 191: Can we have each of these cases to be there own test case? We can get 4 test cases from just this single test.

registrar/apps/api/tests/test_serializers.py
```diff
@@ -36,21 +36,21 @@ def setUp(self):
             PROGRAM_CACHE_KEY_TPL.format(uuid=self.program.discovery_uuid),
             {
                 'title': 'test-program',
-                'type': 'masters',
+                'type': 'Masters',
```
schenedx, line 39: Why do we need this change? I thought the type should be lower cased value coming from discovery?

schenedx (changes requested):
Left a few comments.

registrar/apps/api/tests/test_serializers.py
michaelroytman: I made this change to match the condition used to determine whether a program has enrollments enabled [here](https://github.com/edx/registrar/blob/master/registrar/apps/core/proxies.py#L186). This test program was returning `False` for `is_enrollment_enabled` despite being specified as Master's program for this reason.

The value for `type` coming from Discovery is upper-cased "Masters". I verified this by calling the Discovery API for a Master's program. The `slug`, however, is lower-cased "masters". Also, the [serializer](https://github.com/edx/course-discovery/blob/bc4774182ff3040bfd54501584021516e8856694/course_discovery/apps/api/serializers.py#L1370) used for the endpoint we're calling serializes the `type` attribute based on the `name` field on the `ProgramType` model, which is upper-cased.

registrar/apps/api/serializers.py
```diff
@@ -41,12 +41,8 @@ def get_permissions(self, program):
         """
         request = self.context.get('request')
         if request and hasattr(request, 'user'):
-            user = request.user
-            user_permissions = get_user_api_permissions(
-                user, program
-            ).union(
-                get_user_api_permissions(user, program.managing_organization)
-            )
+            user_permissions = get_effective_user_program_api_permissions(request.user, program)
```
kdmccormick, line 44: :tada: What a nice clean-up.

registrar/apps/core/utils.py
```diff
@@ -26,12 +28,48 @@ def get_user_organizations(user):
...
+    """
+    If a program does not have enrollments enabled, remove the user's
+    permissions to read or write enrollments for that program.
+    """
+    if not program.is_enrollment_enabled:
+        for perm in ENROLLMENT_PERMISSIONS_LIST:
+            if perm in user_permissions:
+                user_permissions.remove(perm)
+
+    return user_permissions
```
kdmccormick: Style nit:

Consider writing this functionally; i.e., return a new set instead of modifying the parameter. Since you're using sets, Python gives you operators that may make it really natural and concise, like `|` for union, `&` for intersect, and `-` for difference. You may even be able to inline this function

registrar/apps/core/utils.py
```diff
@@ -26,12 +28,48 @@ def get_user_organizations(user):
...
 
+def get_effective_user_program_api_permissions(user, program):
+    """
+    Returns a set of all APIPermissions granted to the user on a
+    program  either in the context of a program or an organization.
+    This includes permissions granted though a global permission or role.
+
+    This will filter out APIPermissions for the program that are not valid
+    for the program. Currently, this only removes a user's read and/or write
+    enrollments permissions for a program that does not have enrollments enabled.
```
kdmccormick, line 39: Nice exposition.

registrar/apps/api/tests/test_serializers.py
```diff
@@ -36,21 +36,21 @@ def setUp(self):
...
+        for permissions.
         """
-        fake_get_perms.side_effect = [
-            set([APIWriteEnrollmentsPermission]),  # mocked permissions on program
-            set([APIReadMetadataPermission]),  # mocked permissions on org
-        ]
+        fake_get_perms.return_value = set([
+            APIWriteEnrollmentsPermission,
+            APIReadMetadataPermission,
+        ])
```
kdmccormick:
```suggestion
        fake_get_perms.return_value = {
            APIWriteEnrollmentsPermission,
            APIReadMetadataPermission,
        }
```
Style nit: Python has a special `set` literal syntax.

registrar/apps/core/tests/test_utils.py
kdmccormick: +1 (to Simon's comment about making this 4 test functions)

registrar/apps/core/tests/test_utils.py
```diff
@@ -122,6 +129,91 @@ def test_get_api_permissions(self):
...
+            APIReadEnrollmentsPermission,
+            APIWriteEnrollmentsPermission,
+        ]))
+
+        # read/write enrollments permissions should be removed for a program that does not have
+        # enrollments enabled
+        perms = get_effective_user_program_api_permissions(self.user1, self.non_masters_program)
+        self.assertSetEqual(perms, set([
+            APIReadMetadataPermission,
+        ]))
```
kdmccormick: Style nit:

All of the `self.assert<Type>Equal` methods [are automatically called by `self.assertEqual`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertSetEqual), so you can generally just call that.

Also, now that edX is standardized on `pytest`, the plain old `assert perms == {APIReadMetadataPermission}`  works and will print a helpful message if the test fails.

All of that stuff is preference-based, so if you like how `assertSetEqual` implies to the reader what types it is comparing, then by all means stick with that.

kdmccormick (approved):
Looks great!

I agree with Simon's comment about splitting up the test function, so I'm approving with the assumption that you'll address that.

I left 3 style nits. Up to you whether you address those; in my mind they shouldn't block merging this.

schenedx (approved):
👍

PR #299, opened 2020-04-16
BOM-1525 : Added Python 3.8 for testing

mzulqarnain1:
Relevant JIRA : [https://openedx.atlassian.net/browse/BOM-1525](https://openedx.atlassian.net/browse/BOM-1525)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,11 +1,12 @@
 [tox]
-envlist = python3.6-django{22}
+envlist = py{35,38}-django{22,30}
 skipsdist = true
 
 [testenv]
 passenv = *
 deps =
     django22: -r requirements/django.txt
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -1,11 +1,12 @@
 [tox]
-envlist = python3.6-django{22}
+envlist = py{35,38}-django{22,30}
```
jmbowman: I think registrar is the one service we have that actually uses 3.6 in production, so this should be `py{36,38}` instead.

jmbowman (commented):
Looks good other than the noted point.  It's great to see that things mostly seem to just work under 3.8.

jmbowman (approved):

PR #301, opened 2020-04-23
Add course_staff field to course enrollment csv upload

how015:
MST-189

### DIFF
--- a/registrar/apps/api/v1/tests/test_views.py
+++ b/registrar/apps/api/v1/tests/test_views.py
@@ -2436,12 +2436,107 @@ class CourseEnrollmentUploadTest(EnrollmentUploadMixin, S3MockMixin, RegistrarAP
     event = 'registrar.v1.upload_course_enrollments'
     csv_headers = ('student_key', 'course_id', 'status')
 
-    def build_enrollment(self, status, student_key=None, course_id=None):
-        return {
+    # pylint: disable=missing-function-docstring
+    def build_enrollment(self, status, student_key=None, course_id=None, course_staff=None):
+        enrollment = {
             'status': status,
             'student_key': student_key or uuid.uuid4().hex[0:10],
             'course_id': course_id or uuid.uuid4().hex[0:10],
         }
+        if course_staff is not None:
+            enrollment['course_staff'] = course_staff
+
+        return enrollment
+
+    @mock.patch.object(CourseRunEnrollmentUploadView, 'task_fn', _succeeding_job)
+    def test_enrollment_upload_success_with_course_staff(self):
+        self.csv_headers = self.csv_headers + ('course_staff',)
+        enrollments = [
+            self.build_enrollment('enrolled', '001', None, 'True'),
+            self.build_enrollment('pending', '002', None, 'False'),
+            self.build_enrollment('enrolled', '003', None, ''),
+        ]
+
+        with self.assert_tracking(
+                user=self.stem_admin,
+                program_key=self.cs_program.key,
+                status_code=202
+        ):
+            with activate_waffle_flag('enable_course_role_management', self.stem_admin_group):
+                upload_response = self._upload_enrollments(enrollments)
+
+        self.assertEqual(upload_response.status_code, 202)
+
+        job_response = self.get(upload_response.data['job_url'], self.stem_admin)
+        self.assertEqual(job_response.status_code, 200)
+        self.assertEqual(job_response.data['state'], 'Succeeded')
+
+        filestore = get_enrollment_uploads_filestore()
+        retrieved = filestore.retrieve('/{}/{}.json'.format(
+            'uploads',
+            upload_response.data['job_id'],
+        ))
+        self.assertEqual(json.loads(retrieved), enrollments)
+
+    @mock.patch.object(CourseRunEnrollmentUploadView, 'task_fn', _succeeding_job)
+    @override_flag('enable_course_role_management', active=False)
+    def test_enrollment_upload_success_waffle_off_course_staff_included(self):
+        enrollments = [{'course_id': 'ac2102fb65', 'status': 'enrolled', 'student_key': '001', 'course_staff': 'True'},
+                       {'course_id': '12a8bf290c', 'status': 'pending', 'student_key': '002', 'course_staff': 'False'},
+                       {'course_id': 'e7826f975e', 'status': 'enrolled', 'student_key': '003', 'course_staff': ''}]
+
+        expected_enrollments = [{'course_id': 'ac2102fb65', 'status': 'enrolled', 'student_key': '001'},
+                                {'course_id': '12a8bf290c', 'status': 'pending', 'student_key': '002'},
+                                {'course_id': 'e7826f975e', 'status': 'enrolled', 'student_key': '003'}]
+
+        with self.assert_tracking(
+                user=self.stem_admin,
+                program_key=self.cs_program.key,
+                status_code=202
+        ):
+            upload_response = self._upload_enrollments(enrollments)
+
+        self.assertEqual(upload_response.status_code, 202)
+
+        job_response = self.get(upload_response.data['job_url'], self.stem_admin)
+        self.assertEqual(job_response.status_code, 200)
+        self.assertEqual(job_response.data['state'], 'Succeeded')
+
+        filestore = get_enrollment_uploads_filestore()
+        retrieved = filestore.retrieve('/{}/{}.json'.format(
+            'uploads',
+            upload_response.data['job_id'],
+        ))
+        self.assertEqual(json.loads(retrieved), expected_enrollments)
+
+    @mock.patch.object(CourseRunEnrollmentUploadView, 'task_fn', _succeeding_job)
+    def test_enrollment_upload_success_waffle_on_no_course_staff(self):
+        enrollments = [
+            self.build_enrollment('enrolled', '001'),
+            self.build_enrollment('pending', '002'),
+            self.build_enrollment('enrolled', '003'),
+        ]
+
+        with self.assert_tracking(
+                user=self.stem_admin,
+                program_key=self.cs_program.key,
+                status_code=202
+        ):
+            with activate_waffle_flag('enable_course_role_management', self.stem_admin_group):
+                upload_response = self._upload_enrollments(enrollments)
+
+        self.assertEqual(upload_response.status_code, 202)
+
+        job_response = self.get(upload_response.data['job_url'], self.stem_admin)
+        self.assertEqual(job_response.status_code, 200)
+        self.assertEqual(job_response.data['state'], 'Succeeded')
+
+        filestore = get_enrollment_uploads_filestore()
+        retrieved = filestore.retrieve('/{}/{}.json'.format(
+            'uploads',
+            upload_response.data['job_id'],
+        ))
+        self.assertEqual(json.loads(retrieved), enrollments)
 
 
 @ddt.ddt
--- a/registrar/apps/api/v1/views.py
+++ b/registrar/apps/api/v1/views.py
@@ -446,7 +446,11 @@ class EnrollmentUploadView(JobInvokerMixin, APIView):
     """
     parser_classes = (MultiPartParser,)
     field_names = set()  # Override in subclass
+    optional_fields = set()
     task_fn = None  # Override in subclass
+
+    def get_field_names(self, request):
+        return self.field_names
 
     def post(self, request, *args, **kwargs):
         """ POST handler """
@@ -460,7 +464,7 @@ class EnrollmentUploadView(JobInvokerMixin, APIView):
         if is_enrollment_write_blocked(self.program.key):  # pylint: disable=no-member
             return Response('Job already in progress for program', HTTP_409_CONFLICT)
 
-        enrollments = load_records_from_uploaded_csv(csv_file, self.field_names)
+        enrollments = load_records_from_uploaded_csv(csv_file, self.get_field_names(request), self.optional_fields)
         return self.invoke_upload_job(self.task_fn, json.dumps(enrollments), *args, **kwargs)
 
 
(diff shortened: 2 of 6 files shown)


registrar/apps/api/v1/tests/test_views.py
```diff
@@ -2444,6 +2444,98 @@ def build_enrollment(self, status, student_key=None, course_id=None):
...
+
+    @mock.patch.object(ProgramEnrollmentUploadView, 'task_fn', _succeeding_job)
+    @mock.patch.object(CourseRunEnrollmentUploadView, 'task_fn', _succeeding_job)
+    @override_flag('enable_course_role_management', active=False)
+    def test_enrollment_upload_success_ignore_course_staff_when_waffle_disabled(self):
+        enrollments = [{'course_id': 'ac2102fb65', 'status': 'enrolled', 'student_key': '001', 'course_staff': 'True'},
+                       {'course_id': '12a8bf290c', 'status': 'pending', 'student_key': '002', 'course_staff': 'False'},
+                       {'course_id': 'e7826f975e', 'status': 'enrolled', 'student_key': '003', 'course_staff': 'True'}]
+
+        expected_enrollments = [{'course_id': 'ac2102fb65', 'status': 'enrolled', 'student_key': '001'},
```
schenedx, line 2488: #NIT This array can be constructed based on the last `enrollments` array.

```suggestion
        expected_enrollments = [enrollment_item.pop('course_staff') for enrollment_item in enrollments]
```

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -2444,6 +2444,98 @@ def build_enrollment(self, status, student_key=None, course_id=None):
...
+        self.assertEqual(job_response.data['state'], 'Succeeded')
+
+        filestore = get_enrollment_uploads_filestore()
+        retrieved = filestore.retrieve('/{}/{}.json'.format(
+            'uploads',
+            upload_response.data['job_id'],
+        ))
+        self.assertEqual(json.loads(retrieved), expected_enrollments)
+
+    def test_unmatched_row_data(self):
```
schenedx: I say this test needs be executed when the waffle flag is turned on. Please add the waffle flag on patch.

schenedx (changes requested):
Minor comments. This is close.

registrar/apps/api/v1/tests/test_views.py
kdmccormick: Won't that modify the items in `enrollments`?

registrar/apps/api/v1/views.py
```diff
@@ -460,6 +460,9 @@ def post(self, request, *args, **kwargs):
         if is_enrollment_write_blocked(self.program.key):  # pylint: disable=no-member
             return Response('Job already in progress for program', HTTP_409_CONFLICT)
 
+        if isinstance(self, CourseRunEnrollmentUploadView):
```
zacharis278: Instead of needing to typecheck against the class could we do something more like `if self.get_field_names`?

registrar/apps/api/v1/tests/test_views.py
schenedx: @kdmccormick Huh, I didn't realize that. But I think we can do copy then?
Frankly, I am also fine leaving this as it is.

kdmccormick: As-is strikes me as fine. Sometimes I like when tests have their data fully spelled-out instead of created by Python logic. The latter can obscure what the test is doing and hide bugs.

registrar/apps/api/v1/views.py
michaelroytman: I think that would get you an `AttributeError` if `get_field_names` is not defined on the class. I think the following would work.

```
get_field_names = getattr(self, 'get_field_names', None)
if get_field_names:
    self.field_names = get_field_names(request)
```

michaelroytman: That said, there's something about this that makes me uncomfortable. I don't like that the parent class `EnrollmentUploadView` knows about its child class `CourseRunEnrollmentUploadView`. I think that the original intent of the shared `field_names` field no longer applies here because of the necessary change to `CourseRunEnrollmentUploadView`.

A thing that could be better is to define a method `get_field_names(request)` that each subclass would implement and the `post` method could call without needing to worry about type of the subclass. I realize that the `request` argument would go unused in all but the `CourseRunEnrollmentUploadView` case, though. I'm still thinking about this a bit, but I wanted to get my thoughts on the page.

michaelroytman: I just wanted to follow up on this with some additional thoughts; credit to Cale for explaining this approach to me.

I think a good pattern to follow for something like this is the following.

Implement a base implementation of `get_field_names` on the base class.

This default implementation will return `self.fieldnames`, if it exists on the class. If it doesn't, throw an exception or do whatever is appropriate.

Classes that are able to statically define the fieldnames they care about, like `ProgramEnrollmentUploadView`, will do so.

Classes that are unable to statically define the fieldnames they care about, like `CourseRunEnrollmentUploadView`, will overwrite the implementation of `get_fieldnames` to do what it is they need.

And example of this kind of approach can be found in Django's [get_queryset].(https://docs.djangoproject.com/en/3.0/ref/class-based-views/mixins-single-object/#django.views.generic.detail.SingleObjectMixin.get_queryset) method.

registrar/apps/api/v1/tests/test_views.py
```diff
@@ -2444,6 +2444,126 @@ def build_enrollment(self, status, student_key=None, course_id=None):
         }
 
 
+class CourseEnrollmentUploadCourseStaffTest(CourseEnrollmentUploadTest):
+    """ Test course enrollment csv upload """
+    csv_headers = ('student_key', 'course_id', 'status', 'course_staff')
+
+    # pylint: disable=arguments-differ
+    def build_enrollment(self, status, student_key=None, course_staff='', course_id=None):
```
zacharis278: this is basically the exact same function as CourseEnrollmentUploadTest.build_enrollment with an extra parameter.  I think we can just add an optional course_staff argument there to avoid repeating code.

registrar/apps/enrollments/tests/test_tasks.py
```diff
@@ -387,6 +388,37 @@ def test_success_status(self, any_successes, any_failures, expected_code_str):
...
+    @ddt.data(
+        (True, False, "200"),
+        (True, True, "207"),
+        (False, True, "422"),
+    )
+    @ddt.unpack
+    @patch.object(
+        DiscoveryProgram, 'get_course_key', lambda _self, x: x + '-internal'
+    )
+    def test_success_status_course_staff_included(self, any_successes, any_failures, expected_code_str):
```
zacharis278: IMO re-testing the status codes here is redundant.  I think we only need one successful run of this to validate the course_staff inclusion.

A gap here is that `mock_write_enrollments` doesn't do anything with the passed in `course_staff` value so we aren't really validating that it is included in the JSON payload sent to LMS.  One way to handle this is using `assert_called_with` on the mock object.  See https://docs.python.org/3/library/unittest.mock.html

registrar/apps/core/tests/test_utils.py
```diff
@@ -290,7 +290,7 @@ class LoadRecordsFromCSVStringTests(TestCase):
         "\r\n"
     )
     csv = csv_fmt.format(pepper_is_vegetarian='true')
-    bad_csv = csv_fmt.format(pepper_is_vegetarian='')  # Empty value
+    csv_with_null = csv_fmt.format(pepper_is_vegetarian='')  # Empty value
```
schenedx: #NIT Better variable name for this is `csv_with_empty`

schenedx (commented):
Don't really have other comments besides what Zach has commented. This is definitely more complicated than originally thought. Good job for all those tests and making sure all your test cases are covered.

registrar/apps/api/v1/tests/test_views.py
how015: csv_headers is the reason why I create the new class because csv_headers is also referenced in the parent class of EnrollmentUploadMixin and therefore we can't define two different csv_headers in the same class

zacharis278: I think a simple way to solve that would be to use a getter for CSV headers instead of a static value similar to what we did with the view classes.  e.g. `def get_csv_headers(course_staff_enabled=False):`

registrar/apps/api/v1/views.py
```diff
@@ -460,7 +464,7 @@ def post(self, request, *args, **kwargs):
         if is_enrollment_write_blocked(self.program.key):  # pylint: disable=no-member
             return Response('Job already in progress for program', HTTP_409_CONFLICT)
 
-        enrollments = load_records_from_uploaded_csv(csv_file, self.field_names)
+        enrollments = load_records_from_uploaded_csv(csv_file, self.get_field_names(request), self.optional_fields)
```
how015, line 467: optional_fields=frozenset()

OR

optional_fields=None
(and then in the function)
optional_fields = optional_fields or set()

registrar/apps/core/utils.py
```diff
@@ -124,7 +124,7 @@ def fieldnames(self):
         return [field_name.strip().lower() for field_name in raw_field_names]
 
 
-def load_records_from_uploaded_csv(csv_file, field_names):
+def load_records_from_uploaded_csv(csv_file, field_names, optional_fields=set()):
```
how015: optional_fields=frozenset()

OR

optional_fields=None
(and then in the function)
optional_fields = optional_fields or set()

registrar/apps/core/utils.py
```diff
@@ -133,10 +133,10 @@ def load_records_from_uploaded_csv(csv_file, field_names):
         field_names (set[str])
     """
     contents = csv_file.read().decode('utf-8')
-    return load_records_from_csv(contents, field_names)
+    return load_records_from_csv(contents, field_names, optional_fields)
 
 
-def load_records_from_csv(csv_string, field_names):
+def load_records_from_csv(csv_string, field_names, optional_fields=set()):
```
kdmccormick:
```suggestion
def load_records_from_csv(csv_string, field_names, optional_fields=frozenset()):
```

registrar/apps/core/utils.py
kdmccormick:
```suggestion
def load_records_from_uploaded_csv(csv_file, field_names, optional_fields=frozenset()):
```

registrar/apps/core/utils.py
```diff
@@ -153,14 +153,16 @@ def load_records_from_csv(csv_string, field_names):
     """
     row_iter = csv_string.splitlines()
     reader = StrippedLowercaseFieldNamesDictReader(row_iter)
-    missing_fields = field_names - set(reader.fieldnames)
+    required_fields = field_names - set(optional_fields)
```
kdmccormick, line 156: (ALTERNATIVE SOLUTION)

If using `frozenset` doesn't work...

Do `def load_records_from_csv(csv_string, field_names, optional_fields=None):` in the function signatures,
and then here, do:
```suggestion
    required_fields = field_names - (optional_fields or set())
```

registrar/apps/api/v1/tests/test_views.py
how015: @Zacharis278 would sync with you further on this, tend to think make it separate class may be more adequate.

how015: CourseEnrollmentUploadCourseStaffTest and CourseEnrollmentUploadTest will run the basic tests from EnrollmentUploadMixin with their own csv_headers.

registrar/apps/api/v1/views.py
```diff
@@ -482,10 +486,19 @@ class CourseRunEnrollmentUploadView(EnrollmentMixin, CourseSpecificViewMixin, En
...
     task_fn = write_course_run_enrollments
     event_method_map = {'POST': 'registrar.{api_version}.upload_course_enrollments'}
     event_parameter_map = {'program_key': 'program_key'}
+    optional_fields = ('course_staff',)
+
+    def get_field_names(self, request):
+        """
+        Get field names based on waffle flag
+        """
+        if waffle.flag_is_active(request, 'enable_course_role_management'):
```
matthugs, line 499: How long will we need this waffle flag? Do we have a ticket in JIRA tracking when it can be removed, assuming it's a temporary stage of a roll-out plan? If so, please include a comment referencing that ticket, and include details in the ticket about what needs to be true before this waffle flag can be removed.

registrar/apps/enrollments/tests/test_tasks.py
```diff
@@ -339,17 +339,19 @@ def spawn_task(self, program_key=None, **kwargs):
             task_id=self.job_id
         )
 
-    def mock_write_enrollments(self, any_successes, any_failures):
+    def mock_write_enrollments(self, any_successes, any_failures, expected_enrolls_by_course_key=()):
```
matthugs: it looks like code underneath this assumes that `expected_enrolls_by_course_key` is a dict

matthugs (commented):
(would approve but don't want to do that while travis is not passing)

registrar/apps/enrollments/tests/test_tasks.py
kdmccormick: @how015 If setting the default param value to `{}` gives you a "dangerous default value" Pylint warning, then I suggest setting it to `None`. Then, in the body of the function, you can write:
```python
expected_enrolls_by_course_key = expected_enrolls_by_course_key or {}
```

registrar/apps/api/v1/views.py
how015: MST-190 is the ticket

matthugs: Could you include that ticket number in a line comment above the waffle flag check?

matthugs (approved):

PR #307, opened 2020-05-14
Increase Discovery cache timeout to 24 hours

kdmccormick:
```
It was previously only 2 minutes due to belief
that the delay between modifying an external
course key in Publisher and seeing the result in
the Registrar API needed to be low.

However, the performance issues on Program Console
are currently a larger concern. We believe that
increasing this timeout will work as a stop-gap
to optimize the /api/v2/programs/ endpoint, which
Program Console relies upon for the initial page load.

MST-255
```
@edx/masters-devs-cosmonauts 
https://openedx.atlassian.net/browse/MST-255

This also fix an unrelated deprecation warning coming from Faker (which we use to make test data).


schenedx (approved):
👍

kdmccormick:
Build passed; Travis is just failing to report it.

https://travis-ci.org/github/edx/registrar/builds/686991392

PR #314, opened 2020-06-08
Validate that Organization/Program keys are slugs

kdmccormick:
```
When they are set otherwise, it can cause unexpected
404s in the API due to our URL patterns, which assume
the `key` fields to be slugs.

Resulted from PROD-1669
```
This validation will prevent program/organization creators from setting invalid slugs via the models' Django Admin forms.

Some follow-up from https://openedx.atlassian.net/browse/PROD-1669

@edx/masters-devs


schenedx (approved):
👍

PR #316, opened 2020-06-12
Use UTF-8 Encoding to open the API description file

schenedx:
@edx/masters-devs-cosmonauts Please review
I needed to specify the encoding on this file open for sandbox. I don't know why but this shouldn't hurt on other environments as well.

### DIFF
--- a/registrar/urls.py
+++ b/registrar/urls.py
@@ -34,7 +34,7 @@ admin.autodiscover()
 
 app_name = 'registrar'
 api_description_path = os.path.join(settings.PROJECT_ROOT, 'static', 'api-description.html')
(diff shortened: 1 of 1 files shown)


kdmccormick (approved):
Ha, I think you hit the same issue I did a year ago: https://github.com/edx/registrar/pull/136

A non-ascii character must have crept into the file again. Your solution should make us tolerant to that going forward, which is good.

I believe that UTF-8 is the default encoding in most environments, which is why it isn't an issue usually. I think the sandbox environment is weirdly configured to have ASCII as the default encoding :shrug: Would be good to fix that eventually, probably.

PR #317, opened 2020-06-12
Improve program listing performance (Part 1)

kdmccormick:
In its acceptance criteria, [MST-191](https://openedx.atlassian.net/browse/MST-191) lists three approaches to speeding up the `/api/v1/programs` endpoint:
1. Reduce the number of programs we request from Discovery.
2. Reduce the number of SQL queries necessary to check a user’s permission.
3. Then, either batch-load Discovery programs or create a new ticket to do so.

This PR handles the **first approach**, as well as some additional refactoring. A subsequent PR will handle the second approach, and then we will evaluate the necessity of doing the third approach in this ticket vs. creating a new ticket for it.

**Incidental commits:**
* Remove an outdated TODO comment
* Silence a pylint warning about the `debug_toolbar` import in a way that does not message "useless suppression" during local development.
* Change ProgramFactory and OrganizationFactory to generate predictable sequential UUIDs to make test failures easier to diagnose.
* Change APIPermissions to be objects instead of classes, which is I think is a more idiomatic use of Python objects.

**Interesting commits:**
* Add a failing test that demonstrates an issue: *Any and all calls to the program listing endpoint will query the Discovery cache for every single program in the database*.
* Fix said issue. Change the interface to the Discovery cache to be more explicit, making it less likely to accidentally load up every single program in the database from Discovery.
  * Historical note: I originally developed the proxy-model-based DiscoveryProgram cache in hopes that it would make working with programs in Registrar easier, but I realized now that the "magic-ness" of it hurt readability and made it easier to run into performance gotchas.

@edx/masters-devs-cosmonauts 
https://openedx.atlassian.net/browse/MST-191


registrar/apps/api/constants.py
```diff
@@ -14,12 +14,7 @@
 
 # To be deprecated with upcoming program manager changes (01/2020)
 LEGACY_PERMISSION_QUERY_PARAMS = {
-    'metadata': perms.APIReadMetadataPermission,
```
zacharis278, line 17: afaik we can just delete this entirely, program-console doesn't use these query params anymore

schenedx (approved):
👍 Thank you for the refactoring.

registrar/apps/api/constants.py
kdmccormick: Good call :+1:

zacharis278:
Refactor looks good 🚀

zacharis278 (approved):

PR #319, opened 2020-06-15
Improve program listing performance (Part 2)

kdmccormick:
This PR consolidates all authorization-related logic into `registrar/apps/core/auth_checks.py` and optimizes them. Previously, authorization logic was scattered throughout views, mixins, and utility functions.

As part of that goal, it breaks apart the `AuthMixin` which we previously used in most of our views. By trying to generically handle our compelx auth system for all our API views, `AuthMixin` was getting hard to reason about.

Furthermore, this PR breaks `core/utils.py` into `core/csv_utils.py` and `core/auth_checks.py` to make it more obvious where Auth-related functions live.

Commits:
1. (unrelated) Make it so `make validate` doesn't try to open a browser to show coverage.
2. (unrelated) Re-enable pylinting for `scripts/` directly.
3. Add a failing query count assertion to `get_user_organizations`, which is resolved in the next commit.
4. The meat of this PR.

As an informal improvement metric, calls to `/api/v2/programs?...` in `test_program_filters` used to execute ~25-40 queries per request. That's down to ~16-20 queries. (I didn't do any analysis of the query vs. program growth rate, but I expect that it's now constant where before it was linear)

[Pull request for Part 1](https://github.com/edx/registrar/pull/317) of the program listing improvement..

@edx/masters-devs-cosmonauts 
https://openedx.atlassian.net/browse/MST-191


registrar/apps/api/mixins.py
```diff
@@ -58,7 +58,27 @@ def add_tracking_data(self, **kwargs):
         """
         Add fields to the tracking event.
         """
-        self._extra_tracking_data.update(kwargs)
+        data = kwargs.copy()
```
schenedx: I think it's good to wrap the following 6 lines into a function. Make the purpose of this code easier to understand.

schenedx (approved):
Only a NIT comment. Thank you for the refactoring. The auth_checks.py function is now exactly how we discussed it should look like. Really appreciate the code refactoring.

registrar/apps/api/mixins.py
kdmccormick: Good point. Agreed.

kdmccormick: @schenedx Addressed.

PR #323, opened 2020-06-23
Empty test PR

kdmccormick:
Confirming changes made to Travis config in #322 .


kdmccormick:
Travis looks good. Realizing now that this also will make PRs build a bit faster, because you're waiting for one 3-6 minute process to complete and not two of them.

Looks like Codecov still reports twice, but that's because it's configured for both `patch` and `project`. The former is redundant with the latter at 100%, but I'll leave it because it doesn't affect performance.

PR #335, opened 2020-08-04
WIP: Add dockerfile and github action

katenels:
This will allow us to build Docker images, as well as publish Docker images to Dockerhub.


manage.py
```diff
@@ -8,7 +8,7 @@
 import sys
 
 if __name__ == "__main__":
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "registrar.settings.devstack")
+    os.environ["DJANGO_SETTINGS_MODULE"] = "registrar.settings.devstack"
```
adzuci, line 11: I don't think you need this change.

PR #336, opened 2020-08-07
django3.0 compatible

awais786:
**NOTE: Merge it after registrar deployment on py3.8 in production** 

executing pyupgrade on all .py files.
run make upgrade from py38 env.

## Description


UsamaSadiq (approved):

mraarif (approved):

mzulqarnain1 (approved):

registrar/apps/core/apps.py
```diff
@@ -22,9 +22,15 @@ def ready(self):
...
-        from registrar.apps.core.signals import handle_user_post_save       # pylint: disable=import-outside-toplevel
+        from django.db.models.signals import (  # pylint: disable=import-outside-toplevel
+            post_save,
+            pre_migrate,
+        )
+
+        from registrar.apps.core.models import \
+            User  # pylint: disable=import-outside-toplevel
+        from registrar.apps.core.signals import \
+            handle_user_post_save  # pylint: disable=import-outside-toplevel
```
jmbowman: Hmm, was this all done by isort?  We may want to configure it to use 120 as the line length as we do in edx-platform and elsewhere.  Configuration for that is at https://github.com/edx/registrar/blob/master/setup.cfg#L8

jmbowman (commented):
In addition to the one point I noted, remember that this shouldn't merge until at least a couple of days after registrar deploys to Python 3.8 in production.  Also, there are enough small changes here that we should make sure the owning team is ok with the result.

registrar/apps/core/apps.py
awais786: Change is fixed by `isort` configuring to 120.

PR #338, opened 2020-08-10
UserGroup model

zacharis278:
@edx/masters-devs-cosmonauts 

### [MST-298](https://openedx.atlassian.net/browse/MST-298)

Adds a new model UserGroup that is really just a 1:1 proxy to User.  This model has it's own admin page that only exposes a user's assigned groups.  We can grant access to UserGroup but not User through a globally defined django group.

Looking for feedback since under-the-hood UserGroup is really just User and we can't just drop the undesirable fields off. I could override create/update to throw errors on a hard-coded set of fields but that felt kludgy. Ultimately, this still seemed more elegant than complex Admin logic for User which is another option.


zacharis278:
This is what support staff would see:

image

schenedx (commented):
How do we manage the permissions for non-super user staff to gain access to this proxy model? Do we need to create a new permission group?

zacharis278:
@schenedx yeah we already have the _Global_ProgramManagement group that gives support everything they need right now (except this).  So, we can just add "update UserGroup" to it they will have access.

schenedx (approved):
👍

kdmccormick:
Saw this via a wiki page update notification. Nice simple solution :+1:

PR #340, opened 2020-08-12
Add Dockerfile and Github Action

katenels:
This will allow us to build docker images, as well as publish docker images to dockerhub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,55 @@
+FROM ubuntu:xenial as app
+
+# System requirements.
+RUN apt-get update && apt-get upgrade -qy
+RUN apt-get install -qy \
+	git-core \
+	language-pack-en \
+	python3.5 \
+	python3-pip \
+	python3.5-dev \
+	libmysqlclient-dev \
+	libssl-dev
+RUN pip3 install --upgrade pip setuptools
+RUN rm -rf /var/lib/apt/lists/*
+
+# Python is Python3.
+RUN ln -s /usr/bin/pip3 /usr/bin/pip
+RUN ln -s /usr/bin/python3 /usr/bin/python
+
+# Use UTF-8.
+RUN locale-gen en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US:en
+ENV LC_ALL en_US.UTF-8
+ENV DJANGO_SETTINGS_MODULE registrar.settings.production
+
+RUN mkdir -p /edx/app/registrar
+
+# Expose ports.
+EXPOSE 18734
+EXPOSE 18735
+
+RUN useradd -m --shell /bin/false app
+
+# Working directory will be root of repo.
+WORKDIR /edx/app/registrar
+
+# Copy just Python requirements & install them.
+COPY requirements/ /edx/app/registrar/requirements/
+COPY Makefile /edx/app/registrar/
+RUN make production-requirements
+
+# Code is owned by root so it cannot be modified by the application user.
+# So we copy it before changing users.
+USER app
+
+# Gunicorn 19 does not log to stdout or stderr by default. Once we are past gunicorn 19, the logging to STDOUT need not be specified.
+CMD ["gunicorn", "--workers=2", "--name", "registrar", "-c", "/edx/app/registrar/registrar/docker_gunicorn_configuration.py", "--log-file", "-", "--max-requests=1000", "registrar.wsgi:application"]
+
+# After the requirements so changes to the code will not bust the image cache
+COPY . /edx/app/registrar
+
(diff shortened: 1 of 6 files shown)


adzuci (commented):
This looks good in general, but I think we need to resolve the codecov warnings before merging, cc @jdmulloy

registrar/docker_gunicorn_configuration.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def close_all_caches():
+    """
+    Closes the cache so that newly forked workers cannot accidentally
+    share the socket with the processes they were forked from.
+    Prevents a race condition in which one worker could get a cache response
+    intended for another worker.
+
+    Safe for 1.4 and 1.8.
```
matthugs: I was curious what version numbers we were referring to here. [Based on this](https://github.com/edx/configuration/commit/fa033cc0d57d62d0471e499df3aa5ff1aabcfe24) it looks like these were django versions. We definitely don't use either of those versions of django and haven't for a long time, so I wonder whether or not we need this still?

ohnickmoy:
To resolve those warnings down the line, you can place unittests [in the apps/core/tests part of this repo.](https://github.com/edx/registrar/tree/master/registrar/apps/core/tests)

ohnickmoy (approved):
To resolve those warnings down the line, you can place unittests [in the apps/core/tests part of this repo.](https://github.com/edx/registrar/tree/master/registrar/apps/core/tests). 

unit test docs for python can be found [here](https://docs.python.org/3/library/unittest.html)

registrar/docker_gunicorn_configuration.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Args:
+        worker (Worker): The worker to process the request
+        req (Request): The request to be processed
+
+    Returns:
+        None
+    """
+    worker.log.info("%s %s" % (req.method, req.path))
+
```
ohnickmoy, line 26: @adzuci, is this just printing out log statements, does the req needs to be mocked? https://docs.python.org/3/library/unittest.mock.html

ohnickmoy (approved):
To resolve those warnings down the line, you can place unittests [in the apps/core/tests part of this repo.](https://github.com/edx/registrar/tree/master/registrar/apps/core/tests). 

unit test docs for python can be found [here](https://docs.python.org/3/library/unittest.html)

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+WORKDIR /edx/app/registrar
+
+# Copy just Python requirements & install them.
+COPY requirements/ /edx/app/registrar/requirements/
+COPY Makefile /edx/app/registrar/
+RUN make production-requirements
+
+# Code is owned by root so it cannot be modified by the application user.
+# So we copy it before changing users.
+USER app
```
kdmccormick: Seems like we copy the code later. Can you fix the comment or fix the order of the commands?

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+# Python is Python3.
+RUN ln -s /usr/bin/pip3 /usr/bin/pip
+RUN ln -s /usr/bin/python3 /usr/bin/python
+
+# Use UTF-8.
+RUN locale-gen en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US:en
+ENV LC_ALL en_US.UTF-8
+ENV DJANGO_SETTINGS_MODULE registrar.settings.production
```
kdmccormick, line 27: Style nit: this seems separate from the block of UTF-8-related commands.
```suggestion

ENV DJANGO_SETTINGS_MODULE registrar.settings.production
```

kdmccormick (commented):
Cool; nice work!

A couple comments below. Please also add a `.dockerignore` file to make sure things like `.git/`, `registrar/settings/private.py`, etc. are excluded from the image. [This article](https://codefresh.io/docker-tutorial/not-ignore-dockerignore-2/) does a good job explaining the concept and rationale.

matthugs (approved):

registrar/docker_gunicorn_configuration.py
katenels: I removed that line from the docstring but I don't know enough about this to remove the second part of this method, so I've left it in for now

registrar/docker_gunicorn_configuration.py
adzuci: It's just printing the logs, but to test it we would need to create a sample worker and request object with a method and path which could be passed into `pre_request`.  I don't think we would need to mock the actual request for this method though.

adzuci (approved):
:shipit:

PR #345, opened 2020-08-21
Upgrade to celery4.0

awais786:
PART1.

Python3.8 is not compatible right now with this PR and tests failing.

https://openedx.atlassian.net/browse/BOM-1991

Upgrade to celery==4.0.2 
Upgrade kombu

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -4,11 +4,10 @@
 #
 #    pip-compile --output-file=requirements/base.txt requirements/base.in
 #
-amqp==1.4.9               # via kombu
+amqp==2.6.1               # via kombu
 analytics-python==1.2.9   # via -c requirements/constraints.txt, -r requirements/base.in
-anyjson==0.3.3            # via kombu
-billiard==3.3.0.23        # via celery
-celery==3.1.26.post2      # via -c requirements/constraints.txt, -r requirements/base.in, django-user-tasks
+billiard==3.5.0.5         # via celery
+celery==4.0.2             # via -c requirements/constraints.txt, -r requirements/base.in, django-user-tasks
 certifi==2020.6.20        # via -r requirements/github.txt, requests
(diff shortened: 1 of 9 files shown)


requirements/base.txt
```diff
@@ -23,7 +22,7 @@ django-model-utils==4.0.0  # via -r requirements/base.in, django-user-tasks
 django-mysql==3.8.1       # via -r requirements/base.in
 django-simple-history==2.11.0  # via -r requirements/base.in
 django-storages==1.8      # via -c requirements/constraints.txt, -r requirements/base.in
-django-user-tasks==1.1.0  # via -r requirements/base.in
+django-user-tasks==1.2.0  # via -r requirements/base.in
```
jmbowman, line 25: Ah, this probably doesn't work with celery 4+ yet.  I'll try to fix that early next week, since I already have the context on why it couldn't easily support both versions.

requirements/base.txt
```diff
@@ -38,12 +37,13 @@ edx-opaque-keys==2.1.1    # via edx-drf-extensions
 edx-rest-api-client==1.9.2  # via -c requirements/constraints.txt, -r requirements/base.in
 future==0.18.2            # via pyjwkest
 idna==2.8                 # via -c requirements/constraints.txt, -r requirements/github.txt, requests
+importlib-metadata==1.7.0  # via kombu
 inflection==0.5.0         # via drf-yasg
 itypes==1.2.0             # via coreapi
 jinja2==2.11.2            # via coreschema
-kombu==3.0.37             # via celery
+kombu==4.6.11             # via celery
```
jmbowman, line 44: Hmm, I wonder if this version actually works with celery 4.0.x or they just didn't sufficiently constrain the kombu version in those celery releases.  Would be nice if we could postpone the later celery upgrades until after the Python 3.8 rollout.

PR #347, opened 2020-08-21
MST-316 Track with user_id instead of username to conform to Data Analytics Standards

schenedx:
@brianhw on DE asked me to update the tracking of registrar to be based on user_id instead of username.

@edx/masters-devs-cosmonauts Please review


azarembok (approved):

schenedx:
We want the LMS user_id instead of the registrar native user_id. This PR is not doing the right thing.

PR #351, opened 2020-08-26
Upgrade redis to the latest version

iamsobanjaved:
## Description

Unpinned redis to upgrade to latest version

https://openedx.atlassian.net/browse/BOM-1995


jmbowman (approved):
Looks good to me.  As far as I can tell, redis is only used as the production celery broker here.  I've asked @kdmccormick why it was downgraded in the first place (registrar used to use 3.2.1), let's wait a little before merging to give him time to reply in case I'm missing something.

kdmccormick:
@jmbowman I'm struggling to remember why I pinned it to 2.8 It might have been related to the fact that we pin an old version of Celery in order to use django-user-tasks, but I'm not sure.

In lieu of an obvious reason not to upgrade it, you're probably good to merge this, and then ask @edx/masters-devs-cosmonauts confirm that the Celery tasks still work on Stage.

FWIW, we do use Redis as a broker in Devstack as well.

iamsobanjaved:
@edx/masters-devs-cosmonauts can you confirm that celery tasks are working on stage after this merge?

PR #354, opened 2020-09-01
updated the Dockerfile to use python 3.8

mraarif:
## Description
Update the Dockerfile to use python 3.8 to replace python 3.5
Relevant JIRA issue here: https://openedx.atlassian.net/browse/BOM-2007


mraarif:
I was able to build an image with these changes

Dockerfile
```diff
@@ -1,21 +1,24 @@
...
 RUN rm -rf /var/lib/apt/lists/*
 
-# Python is Python3.
-RUN ln -s /usr/bin/pip3 /usr/bin/pip
-RUN ln -s /usr/bin/python3 /usr/bin/python
+ENV VIRTUAL_ENV=/venv
+RUN python3.8 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+RUN pip install pip==20.2.3 setuptools==50.3.0
```
jmbowman: Could we put these versions in a requirements file, and then do these upgrades in the `piptools` make target before installing `pip-tools.txt`?  This would keep the versions consistent everywhere we use the service, although in that case we should match the versions currently used by the registrar virtualenv created by the `configuration` repo.  So basically add a new first command to that target like `pip install -r requirements/pip.txt`.

My thinking is that we want to start upgrading these more often, which would cause a lot of changes to this file unrelated to how we build the Docker image.  Whereas we already keep all our other frequently updating package versions in the `requirements` directory.

Dockerfile
```diff
@@ -1,21 +1,24 @@
...
+	python3.8-venv \
 	libmysqlclient-dev \
 	libssl-dev
-RUN pip3 install --upgrade pip setuptools
 RUN rm -rf /var/lib/apt/lists/*
 
-# Python is Python3.
-RUN ln -s /usr/bin/pip3 /usr/bin/pip
-RUN ln -s /usr/bin/python3 /usr/bin/python
+ENV VIRTUAL_ENV=/venv
```
jmbowman: This isn't where devstack keeps it, and while this is simpler and easier to remember there are some arguments against it outlined in https://edx-internal.slack.com/archives/CCE8D250D/p1601488733235100 .  In particular:

* We currently try to keep all edX-specific stuff under the `/edx` directory
* Some services need multiple virtualenvs, and we'd rather keep the path pattern consistent between services
* There might be existing code assuming the current virtualenv path

So this would switch to `ENV VIRTUAL_ENV=/edx/app/registrar/venvs/registrar`

mraarif:
I replaced the existing `edxops/registrar` image with new one, ran `make dev.up`, ran `make registrar-shell` and ran `make requirements` that failed on
```ERROR: Exception:
Traceback (most recent call last):
  File "/usr/lib/python3.8/shutil.py", line 788, in move
    os.rename(src, real_dst)
PermissionError: [Errno 13] Permission denied: '/venv/lib/python3.8/site-packages/gevent-20.6.2.dist-info/' -> '/tmp/pip-uninstall-04blskvj'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/venv/lib/python3.8/site-packages/pip/_internal/cli/base_command.py", line 228, in _main
    status = self.run(options, args)
  File "/venv/lib/python3.8/site-packages/pip/_internal/commands/uninstall.py", line 89, in run
    uninstall_pathset = req.uninstall(
  File "/venv/lib/python3.8/site-packages/pip/_internal/req/req_install.py", line 686, in uninstall
    uninstalled_pathset.remove(auto_confirm, verbose)
  File "/venv/lib/python3.8/site-packages/pip/_internal/req/req_uninstall.py", line 394, in remove
    moved.stash(path)
  File "/venv/lib/python3.8/site-packages/pip/_internal/req/req_uninstall.py", line 283, in stash
    renames(path, new_path)
  File "/venv/lib/python3.8/site-packages/pip/_internal/utils/misc.py", line 352, in renames
    shutil.move(old, new)
  File "/usr/lib/python3.8/shutil.py", line 800, in move
    rmtree(src)
  File "/usr/lib/python3.8/shutil.py", line 715, in rmtree
    _rmtree_safe_fd(fd, path, onerror)
  File "/usr/lib/python3.8/shutil.py", line 672, in _rmtree_safe_fd
    onerror(os.unlink, fullname, sys.exc_info())
  File "/usr/lib/python3.8/shutil.py", line 670, in _rmtree_safe_fd
    os.unlink(entry.name, dir_fd=topfd)
PermissionError: [Errno 13] Permission denied: 'RECORD'

**Comment by jmbowman** [comment]
That's because the user switched to `app` after dependencies were installed as `root`.  For this to work in devstack, we'd need a `FROM app as devstack` section that changes the user back to `root`, like we've added in the course-discovery Dockerfile (and then use the image produced for that devstack variant).

**Comment by mraarif** [comment]
As it came up during discussions Registrar is still using configuration repo to build images for deployment, I've opened a PR in configuration repo [configuration#6059](https://github.com/edx/configuration/pull/6059) to switch Registrar to use python 3.8

File: Dockerfile [hunk 3]
```diff
@@ -40,8 +42,14 @@ WORKDIR /edx/app/registrar
...
 COPY Makefile /edx/app/registrar/
+
+RUN pip install -r requirements/pip.txt
 RUN make production-requirements
 
+FROM app as devstack
+USER root
+RUN make devstack-requirements
+
 USER app
```
jmbowman, line 49: I think this needs to be before `from app as devstack`, otherwise the base `latest`-tagged images will switch from `app` to `root`.

jmbowman (commented):
It's been a while since I looked at this, but I think the user change timing issue is the last blocker.

mraarif:
@bseverino , @schenedx I've built a new image with changes suggested by Jeremy, since registrar doesn't have a `docker-compose` how can I test the newly built image, not sure if replacing `edxops/registrar` with this image is how it should be tested.

schenedx:
@mraarif You are correct that registrar do not have it's own standalone docker-compose file. The only docker-compose block is in https://github.com/edx/devstack/blob/master/docker-compose.yml#L333.
I guess you have the following choices:
1. Create your own docker-compose file and also include that in this PR
2. Use the image you built, and upload that to the edxops/registrar image repo. That has the risk of exposing the new image to everyone when we do docker pull.

My question is, what are you looking to test? Is it just running the docker image? If you are able to build the docker image on your computer, why not just test on your own computer?
Or is your question really about how to roll your changes out to rest of us?

mraarif:
@schenedx thanks for the detailed answer, I'm trying to verify if the app inside the container works as expected or as it's working currently when run using the image from `configuration` repo.
I'll go ahead and try to replicate devstack's `docker-compose` for registrar, Once that part is working you guys can test and prioritize when and how would you want to roll out this image to devstack and eventually get it deployed.

Dockerfile
```diff
@@ -40,16 +42,23 @@ WORKDIR /edx/app/registrar
...
 FROM app as newrelic
-RUN pip3 install newrelic
+RUN pip install newrelic
 CMD ["newrelic-admin", "run-program", "gunicorn", "--workers=2", "--name", "registrar", "-c", "/edx/app/registrar/registrar/docker_gunicorn_configuration.py", "--log-file", "-", "--max-requests=1000", "registrar.wsgi:application"]
+
+
+FROM app as devstack
+USER root
+RUN make devstack-requirements
+USER app
```
jmbowman, line 62: Did we decide in other services to just leave the devstack image configured to use `root`?  I thought switching back to `app` prevented any package upgrades during development due to file permission issues.

mraarif: we did similar to this in license-manager: https://github.com/edx/license-manager/pull/115

UsamaSadiq:
Dockerfile has been updated to `Python 3.8` in the PR https://github.com/openedx/registrar/pull/405 so this PR is probably not needed anymore.

openedx-webhooks:
Although this pull request is already closed, I've created [OSPR-6629](https://openedx.atlassian.net/browse/OSPR-6629) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #356, opened 2020-09-08
Add v3 endpoint for program list pagination

alangsto:
## [MST-383](https://openedx.atlassian.net/browse/MST-383)

Added new v3 endpoint to allow users to specify a page number and page size as a query parameter. If no page number or size are specified, there is a default page number of 1 and size of 100.

Most of the tests were copied from the v1 ProgramListView file, but modified to account for the new response format. There are also tests specifically for pagination query parameters. 

The new response format can be found in pagination.py

### DIFF
--- a/registrar/apps/api/v3/pagination.py
+++ b/registrar/apps/api/v3/pagination.py
@@ -1,0 +1,15 @@
+"""Pagination class for API V3"""
+from rest_framework.pagination import PageNumberPagination
+
+
+DEFAULT_PAGE = 1
+DEFAULT_PAGE_SIZE = 100
+
+
+class CustomPagination(PageNumberPagination):
+    """
+    Custom class for paginating objects
+    """
+    page = DEFAULT_PAGE
+    page_size = DEFAULT_PAGE_SIZE
+    page_size_query_param = 'page_size'
--- a/registrar/apps/api/v3/tests/test_views.py
+++ b/registrar/apps/api/v3/tests/test_views.py
@@ -1,0 +1,86 @@
+""" Tests for API views. """
+import ddt
+
+from registrar.apps.api.tests.mixins import AuthRequestMixin
+from registrar.apps.api.v1.tests.test_views import RegistrarAPITestCase
+
+
+@ddt.ddt
+class ViewMethodNotSupportedTests(RegistrarAPITestCase, AuthRequestMixin):
+    """ Tests for the case if user requested a not supported HTTP method """
+
+    api_root = '/api/v3/'
+    TEST_PROGRAM_URL_TPL = 'http://registrar-test-data.edx.org/{key}/'
+    method = 'DELETE'
+    path = 'programs'
+
+    @ddt.data(
+        ('programs', 'ProgramListPaginationView'),
+    )
+    @ddt.unpack
+    def test_not_supported_http_method(self, path, view_name):
+        self.mock_logging.reset_mock()
+        self.path = path
+        response = self.request(self.method, path, self.edx_admin)
+        self.mock_logging.error.assert_called_once_with(
+            'Segment tracking event name not found for request method %s on view %s',
+            self.method,
+            view_name,
+        )
+        self.assertEqual(response.status_code, 405)
+
+
+@ddt.ddt
+class ProgramListPaginationViewTests(RegistrarAPITestCase, AuthRequestMixin):
+    """ Tests for the /api/v3/programs?org={org_key} endpoint with pagination """
+
+    api_root = '/api/v3/'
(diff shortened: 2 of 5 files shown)


registrar/apps/api/v3/tests/test_views.py
```diff
@@ -0,0 +1,641 @@
...
+    @mock.patch.object(Organization.objects, 'get', wraps=Organization.objects.get)
+    def test_org_property_caching(self, get_org_wrapper):
+        # If the 'managing_organization' property is not cached, a single
+        # call to this endpoint would cause multiple Organization queries
+        response = self.get("programs?org=stem-institute", self.stem_admin)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.data['results']), 2)
+        get_org_wrapper.assert_called_once()
+
+    def test_valid_pagination(self):
```
alangsto, line 43: New pagination tests start here

registrar/apps/api/v3/pagination.py
```diff
@@ -0,0 +1,28 @@
...
+class CustomPagination(PageNumberPagination):
+    """
+    Custom class for paginating objects
+    """
+    page = DEFAULT_PAGE
+    page_size = DEFAULT_PAGE_SIZE
+    page_size_query_param = 'page_size'
+
+    def get_paginated_response(self, data):
+        return Response({
```
schenedx: Why do you need the customization on the PageNumberPagination this way?
Why do we need `next` and `previous` wrapped in another level?

registrar/apps/api/v3/tests/test_views.py
```diff
@@ -0,0 +1,584 @@
...
+        self.mock_logging.error.assert_called_once_with(
+            'Segment tracking event name not found for request method %s on view %s',
+            self.method,
+            view_name,
+        )
+        self.assertEqual(response.status_code, 405)
+
+
+@ddt.ddt
+class ProgramListViewTests(RegistrarAPITestCase, AuthRequestMixin):
```
schenedx: Instead of copy and paste the tests here, can't you do refactoring on the original ProgramListViewTests, then inherit those test here?

registrar/apps/api/v3/views.py
```diff
@@ -0,0 +1,62 @@
...
+
+logger = logging.getLogger(__name__)
+
+SCHEMA_COMMON_RESPONSES = {
+    401: 'User is not authenticated.',
+    405: 'HTTP method not support on this path.'
+}
+
+
+@schema_for(
```
schenedx, line 29: Nice. I like how you are using the new doc system now.

schenedx (changes requested):
I am not sure why you need the customized paginator.

registrar/apps/api/v3/pagination.py
alangsto: I agree with your comment that customization can always be added later if necessary, so I removed the custom response.

registrar/apps/api/v3/tests/test_views.py
alangsto: I wasn't quite sure of a good way to refactor and inherit because the response format of this endpoint is different. All of the V1 endpoints look at response.data, but in V3 they need to look at response.data.results. I'll work on changing it!

schenedx: Yep, you can override the function `assert_tracking` to help with that I believe.

alangsto: I could definitely inherit the RegistrarAPITestCase class so I will update the code to include that.

alangsto: @schenedx I've tried to use inheritance for reusing tests, but I'm running into some problems (current code is up to date and reflects these issues). The problem that I'm running into is that by using inheritance, the tests are each called twice. This is causing issues with the build. 

I've found that adding `__test__=False` to the class being inherited fixes the issue of tests being called twice, but that inherited class should be tested normally (for v1 tests). Adding `__test__=False` prevents this normal behavior.

My next thought is to add a base test class which can be inherited for both the v1 and v3 tests, because then `__test__=False` can be added. However, when this is added, ddt.data is ignored in the base test class, and then it would need to be added to both the v1 and v3 test class again. This option seems more complicated than I would have liked, especially because it includes rewriting the v1 testing class, so I feel like copying may be the best option.

Let me know if I was off base with how I was using inheritance, maybe there's something I'm missing

alangsto: @schenedx Just wanted to note that for reusability purposes, adding a base test class is the best option, I just wasn't sure if that would be refactoring the code too much

alangsto:
@schenedx thanks for the feedback! I removed the v1 tests.

schenedx (approved):
👍

PR #360, opened 2020-09-21
Exclude internal endpoints from new api doc

alangsto:
## [MST-429](https://openedx.atlassian.net/browse/MST-429)

Upon loading, the new api-docs page shows an error for an invalid endpoint. This is caused by the two endpoints defined in internal/urls.py, as they both are delete methods and have the same path (parameters do not make a path unique, according to openapi, which is used by api-doc-tools).

Because we wanted to hide these endpoints on the doc anyway, a decorator has been added that will prevent the docs for these endpoints from being generated.


schenedx (approved):
👍

PR #364, opened 2020-09-29
Update celery settings name and added namespace

iamsobanjaved:
## Description

Update celery settings name and added namespace

Relevant JIRA issue: https://openedx.atlassian.net/browse/BOM-2035

### DIFF
--- a/registrar/apps/core/jobs.py
+++ b/registrar/apps/core/jobs.py
@@ -59,7 +59,7 @@ def start_job(user, task_fn, *args, **kwargs):
         [job_id, user.id] + list(args),
         kwargs,
         task_id=job_id,
-        queue=settings.CELERY_DEFAULT_QUEUE,
+        queue=settings.CELERY_TASK_DEFAULT_QUEUE,
     )
     return job_id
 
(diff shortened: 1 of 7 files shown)


registrar/apps/core/jobs.py
```diff
@@ -59,7 +59,7 @@ def start_job(user, task_fn, *args, **kwargs):
         [job_id, user.id] + list(args),
         kwargs,
         task_id=job_id,
-        queue=settings.CELERY_DEFAULT_QUEUE,
+        queue=settings.CELERY_TASK_DEFAULT_QUEUE,
```
jmbowman, line 62: Ah, I'd missed that some of the setting names changed even if you stuck with the all-caps names.  That explains some of the trouble we hit in edx-platform.

registrar/settings/base.py
```diff
@@ -107,46 +107,46 @@
 CELERY_TASK_SERIALIZER = 'json'
 CELERY_RESULT_SERIALIZER = 'json'
 CELERY_ACCEPT_CONTENT = ['json']
-CELERY_MESSAGE_COMPRESSION = 'gzip'
+CELERY_TASK_COMPRESSION = 'gzip'
```
jmbowman, line 110: Do we also want `CELERY_RESULT_COMPRESSION = 'gzip'`?  It looks like the old setting did both.

jmbowman (approved):
One question, otherwise looks good.

awais786 (approved):

PR #366, opened 2020-09-29
Updated registrar sync command

alangsto:
## [MST-434](https://openedx.atlassian.net/browse/MST-434)

@edx/masters-devs-cosmonauts 

The first version of the command did not account for multiple roles within an existing organization group, so logic was included to update only the name of the role, and not the string corresponding to the role name. 

Logic is also included to account for existing organizations that may not have a `ReadOrganizationReports` role associated with it, because in the previous version, organizations that fell under that case would not have this new group/role created.

In order to first replicate this error, I made modifications to the testing file that is included in this PR. Once the additional logic was added, all tests passed with these new organization groups added.

### DIFF
--- a/registrar/apps/core/management/commands/test/test_sync_with_discovery.py
+++ b/registrar/apps/core/management/commands/test/test_sync_with_discovery.py
@@ -12,6 +12,7 @@ from registrar.apps.core.models import (
     ProgramOrganizationGroup,
 )
 from registrar.apps.core.permissions import (
+    OrganizationReadEnrollmentsRole,
     OrganizationReadReportRole,
     ProgramReadReportRole,
 )
@@ -47,10 +48,15 @@ class TestSyncWithDiscoveryCommandBase(TestCase):
             organization=cls.org,
             role=OrganizationReadReportRole.name
         )
-        cls.other_org_read_reports_group = OrganizationGroupFactory(
-            name='{}_ReadOrganizationReports'.format(cls.other_org.key),
+        cls.org_read_enrollments = OrganizationGroupFactory(
+            name='{}_ReadEnrollments'.format(cls.org.key),
+            organization=cls.org,
+            role=OrganizationReadEnrollmentsRole.name
+        )
+        cls.other_org_read_enrollments = OrganizationGroupFactory(
+            name='{}_ReadEnrollments'.format(cls.other_org.key),
             organization=cls.other_org,
-            role=OrganizationReadReportRole.name
+            role=OrganizationReadEnrollmentsRole.name
         )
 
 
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -12,8 +12,10 @@
 # serve to show the default.
 
 import os
+
```
schenedx: #NIT can we not introduce these unnecessary spacing changes?

registrar/apps/core/management/commands/test/test_sync_with_discovery.py
```diff
@@ -47,10 +48,15 @@ def setUpTestData(cls):
...
             role=OrganizationReadReportRole.name
         )
-        cls.other_org_read_reports_group = OrganizationGroupFactory(
-            name='{}_ReadOrganizationReports'.format(cls.other_org.key),
+        cls.org_read_enrollments = OrganizationGroupFactory(
+            name='{}_ReadEnrollments'.format(cls.org.key),
+            organization=cls.org,
+            role=OrganizationReadEnrollmentsRole.name
+        )
+        cls.other_org_read_enrollments = OrganizationGroupFactory(
```
schenedx, line 56: This is not referenced anywhere.  Why do we need the `other_org_read_enrollments group`?

schenedx (commented):
Can you explain why in stage, we had duplicated readReports groups created? And how your PR here fixes that issue on stage?

registrar/apps/core/management/commands/test/test_sync_with_discovery.py
alangsto: In order to replicate what happened on stage, I wanted to test three kinds of orgs: one with two groups (including a read reports group), one with no groups, and one that had existing groups that did not include a read reports group. `other_org_read_enrollments` allows me to test the last case and ensures that existing orgs with no read reports group will have one created.

alangsto:
@schenedx there were duplicate readReports groups due to this function in the original:
```
def update_org_groups(self, updated_orgs):
        """
        Update the existing OrganizationGroups to match the up to date name of the organization
        """
        existing_org_groups = OrganizationGroup.objects.select_related('organization').filter(
            organization__discovery_uuid__in=[org.discovery_uuid for org in updated_orgs]
        )
        org_group_to_update = []
        for org_group in existing_org_groups:
            org_group.name = '{}_ReadOrganizationReports'.format(org_group.organization.key)
            org_group_to_update.append(org_group)

```

The line: `org_group.name = '{}_ReadOrganizationReports'.format(org_group.organization.key)` rewrites every org group name for the org groups that are in the query set, so a Mattx_ReadEnrollments would become Mattx_ReadOrganizationReports.

These two lines (https://github.com/edx/registrar/pull/366/files#diff-7253cbab64da1f3665d25be5d804b31cR169-R170) address this issue.

schenedx (approved):
👍

PR #370, opened 2020-10-02
Update .travis.yml to use ubuntu focal 20.04

bseverino:
### [MST-387](https://openedx.atlassian.net/browse/MST-387)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,10 @@ branches:
   only:
   - master
 
+dist: focal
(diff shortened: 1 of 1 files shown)


zacharis278:
I'm thinking there's no need to run python 3.5 at all at this point?

bseverino:
@Zacharis278 Yeah, I wasn't sure if I should remove it completely. I can update it.

PR #376, opened 2020-10-14
Add option to filter ProgramListView by program_title

bseverino:
### [MST-460](https://openedx.atlassian.net/browse/MST-460)


schenedx (changes requested):
Can you please add another unit test to ensure if the query string doesn't return any programs, the endpoint behavior is expectedly empty?

schenedx (approved):
👍

PR #377, opened 2020-10-19
Programs Sync Fixes

zacharis278:
@edx/masters-devs-cosmonauts 

## [MST-484](https://openedx.atlassian.net/browse/MST-484)

The main change here is that the sync of programs and orgs have been decoupled from the sync of the corresponding groups.  This should simplify the logic needed to deal with states caused by manually creating/modifying these models (which have caused the job to fail).

Added a bit more logging to help track down any other issues that may come up.

One change in here I'm putting up for discussion (https://github.com/edx/registrar/commit/2ef090985c938537824ab46f9d08f62d4eebcdf5) I limited the sync to only the reporting groups. Previously this would alter the name of enrollment groups if and only if the organization name changed. If we're going to enforce rules on naming enrollment groups we should do it holistically and not as a side effect IMO.

### DIFF
--- a/registrar/apps/core/management/commands/sync_with_discovery.py
+++ b/registrar/apps/core/management/commands/sync_with_discovery.py
@@ -31,9 +31,11 @@ class Command(BaseCommand):
         """
         The main logic and entry point of the management command
         """
-        # First make api call to discovery service to get list of organizations available
         self.sync_organizations()
         self.sync_programs(PROGRAM_TYPES_TO_SYNC)
+
+        self.create_org_groups()
+        self.create_program_org_groups()
 
     def sync_organizations(self):
         """
@@ -44,7 +46,6 @@ class Command(BaseCommand):
         discovery_organizations = DiscoveryServiceClient.get_organizations()
         existing_org_dictionary = {}
         orgs_to_create = []
-        org_groups_to_create = []
         orgs_to_update = []
 
         for org in Organization.objects.all():
@@ -60,7 +61,6 @@ class Command(BaseCommand):
                     key=discovery_org.get('key'),
                 )
                 orgs_to_create.append(org)
-                org_groups_to_create.append(org)
                 logger.info('Creating %s', discovery_org.get('key'))
             else:
                 if existing_org.name != discovery_org.get('name') or existing_org.key != discovery_org.get('key'):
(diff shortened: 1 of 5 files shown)


registrar/apps/core/models.py
```diff
@@ -165,10 +165,10 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # Save the value of organization in an attribute, so that when
         # save() is called, we have access to the old value.
-        try:
-            self._initial_organization = self.organization
-        except Organization.DoesNotExist:   # pragma: no cover
-            self._initial_organization = None
+        # try:
```
zacharis278: Looking into another way to solve https://github.com/edx/registrar/pull/32. This block is causing a bunch of repeated queries on the program model. Which, blows up the query counts for this command.

zacharis278: I'm going to ahead and move this fix to another PR since it isn't directly related to the management command itself.

registrar/apps/core/management/commands/sync_with_discovery.py
```diff
@@ -136,90 +128,103 @@ def sync_programs(self, program_types):
...
+            read_reports_groups[(group.granting_organization.key, group.program.key)] = group
+
+        for program in Program.objects.select_related('managing_organization'):
+            name = '{}_{}_ReadProgramReports'.format(
+                program.managing_organization.key,
+                program.key,
             )
+
+            created, updated = (False, False)
+            program_org_group = read_reports_groups.get((program.managing_organization.key, program.key))
```
schenedx, line 207: OK, I can see why your change here will fix for production. It make sense.
That said, I feel like the production issue have the root cause where we have two program with exactly the same name and org, but one is `active` status and one is `retired` status. Should we also no sync programs with `retired` status? Maybe this PR should also include that, when we query programs from discovery, we should only retrieve those `active` programs?

schenedx (approved):
Either way, I think this PR make sense for now. We can optimize later.

registrar/apps/core/management/commands/sync_with_discovery.py
zacharis278: This is a good point about the status. I don't think it's causing the current program that's failing (it only has one `active` record) but it sounds like we should cover it since there is no value in syncing those programs over.

PR #379, opened 2020-10-20
Program/Org Group Performance Fix

zacharis278:
@edx/masters-devs-cosmonauts 

This fixes an n+1 query caused by referencing a model attribute (via FK) in the `__init__` method.  What happens is any query on that model (e.g. .all() or .filter()) will produce an extra SQL query for that mode attribute on every object returned by your original query.

I've moved this logic to a presave hook so it's only hit on save and not every time the Python object is created.

### DIFF
--- a/registrar/apps/core/models.py
+++ b/registrar/apps/core/models.py
@@ -161,15 +161,6 @@ class OrganizationGroup(Group):
         default=perms.OrganizationReadMetadataRole.name,
     )
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        # Save the value of organization in an attribute, so that when
-        # save() is called, we have access to the old value.
-        try:
-            self._initial_organization = self.organization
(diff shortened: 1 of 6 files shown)


registrar/apps/core/models.py
```diff
@@ -161,15 +161,6 @@ class Meta:
         default=perms.OrganizationReadMetadataRole.name,
     )
```
schenedx, line 167: Can you please add some comments in this class to let developer know that there are pre_save signals associated with this model?

registrar/apps/core/models.py
```diff
@@ -230,13 +221,6 @@ class Meta:
         default=perms.ProgramReadMetadataRole.name,
     )
 
-    def __init__(self, *args, **kwargs):
```
schenedx, line 233: Comments about pre_save signal in this model as well.

schenedx (approved):
#NIT suggestion on comments on the model class itself.
LGTM

PR #392, opened 2021-01-26
Removed pylint and edx-lint from constraints.txt

Jawayria:
## Description

TODO: Fix Quality Failures in registrar on latest version of pylint and edx-lint and then unpin pylint from constraints.

JIRA: https://openedx.atlassian.net/browse/BOM-2193

### DIFF
--- a/registrar/apps/api/v1/views.py
+++ b/registrar/apps/api/v1/views.py
@@ -325,7 +325,7 @@ class ProgramEnrollmentView(EnrollmentMixin, JobInvokerMixin, APIView):
             **SCHEMA_COMMON_RESPONSES,
         },
     )
-    def get(self, request, *args, **kwargs):  # pylint: disable=W0613
+    def get(self, request, *args, **kwargs):  # pylint: disable=unused-argument
         """
         Request program enrollment data
 
@@ -360,7 +360,7 @@ class ProgramEnrollmentView(EnrollmentMixin, JobInvokerMixin, APIView):
             **SCHEMA_COMMON_RESPONSES,
         },
     )
-    def post(self, request, program_key):  # pylint: disable=W0613
+    def post(self, request, program_key):  # pylint: disable=unused-argument
         """
         Enroll students in a program
 
@@ -407,7 +407,7 @@ class ProgramEnrollmentView(EnrollmentMixin, JobInvokerMixin, APIView):
             **SCHEMA_COMMON_RESPONSES,
         },
     )
-    def patch(self, request, program_key):  # pylint: disable=W0613
+    def patch(self, request, program_key):  # pylint: disable=unused-argument
         """
         Modify program enrollments
 
@@ -508,7 +508,7 @@ class CourseEnrollmentView(CourseSpecificViewMixin, JobInvokerMixin, EnrollmentM
             **SCHEMA_COMMON_RESPONSES,
         },
     )
-    def get(self, request, *args, **kwargs):  # pylint: disable=W0613
+    def get(self, request, *args, **kwargs):  # pylint: disable=unused-argument
         """
         Request course enrollment data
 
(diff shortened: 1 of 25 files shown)


mzulqarnain1 (approved):

registrar/apps/api/v1/views.py
```diff
@@ -325,7 +325,7 @@ class ProgramEnrollmentView(EnrollmentMixin, JobInvokerMixin, APIView):
             **SCHEMA_COMMON_RESPONSES,
         },
     )
-    def get(self, request, *args, **kwargs):
+    def get(self, request, *args, **kwargs):  # pylint: disable=W0613
```
jmbowman: Could we use the warning name (unused-argument) instead of the number for these so it's more obvious at a glance which warning is being disabled?

registrar/apps/core/apps.py
```diff
@@ -49,6 +54,9 @@ def _disconnect_user_post_save_for_migrations(self, sender, **kwargs):  # pylint
         """
         Handle pre_migrate signal - disconnect User post_save handler.
         """
-        from django.db.models.signals import post_save      # pylint: disable=import-outside-toplevel
-        from registrar.apps.core.models import User     # pylint: disable=import-outside-toplevel
```
jmbowman, line 53: These shouldn't have been long enough to trigger the line length warning; why was this change needed?

registrar/apps/core/models.py
```diff
@@ -180,8 +180,7 @@ def role_object(self):
                 return role
         return None  # pragma: no cover
 
-    # pylint: disable=arguments-differ
-    def save(self, *args, **kwargs):
+    def save(self, *args, **kwargs):  # pylint: disable=W0222
```
jmbowman: Here also, please use the warning name (signature-differs) instead of its number.

registrar/docker_gunicorn_configuration.py
```diff
@@ -39,7 +38,8 @@ def close_all_caches():
         None
     """
     from django.conf import settings  # pylint: disable=import-outside-toplevel
-    from django.core import cache as django_cache  # pylint: disable=import-outside-toplevel
+    from django.core import \
+        cache as django_cache  # pylint: disable=import-outside-toplevel
```
jmbowman: This also didn't seem long enough to require the line split.

requirements/constraints.txt
```diff
@@ -44,3 +40,9 @@ PyYAML==5.1           # https://github.com/edx/registrar/pull/48
 # Constraints required to resolve requirements in docs.in and test.in
 docutils==0.15.2
 idna==2.8
+
+# Latest version requires PyJWT>=2.0.0 but drf-jwt requires PyJWT[crypto]<2.0.0,>=1.5.2
+social-auth-core<4.0.3
+
+# Latest version requires Python 3.6+
+newrelic<6.0
```
jmbowman, line 48: Are we actually still using this repo with Python 3.5 anywhere?  It looks like CI is now 3.8-only.

jmbowman (commented):
Looks good overall, just a few questions and suggestions.

requirements/constraints.txt
mzulqarnain1: The `Dockerfile` is still using 3.5, there is a PR up for that which hasn't been merged yet. https://github.com/edx/registrar/pull/354

mzulqarnain1 (approved):

jmbowman (approved):
This looks good to go, although it might be a good idea to wait until after https://github.com/edx/registrar/pull/354 is merged, just in case there's a change here that would break under 3.5 for the Docker image (since we aren't actually running tests under 3.5 anymore).

UsamaSadiq:
These constraints have been removed in a separate PR so this PR is not needed anymore.

PR #395, opened 2021-02-02
update openedx.yaml to use current best practices

michaelroytman:
## Description

Remove deprecated fields from openedx.yaml, per the [OEP](https://open-edx-proposals.readthedocs.io/en/latest/oep-0002-bp-repo-metadata.html). The ownership document should the single source of truth for ownership information.


schenedx (commented):
Please check if those spaces are indeed your intended changes.

PR #400, opened 2021-03-03
Switch to GitHub CI

alangsto:
## [MST-553](https://openedx.atlassian.net/browse/MST-553)

@edx/masters-devs-cosmonauts 

Update repo to use GitHub Actions instead of Travis


schenedx (approved):
👍 Let's try this.

PR #405, opened 2021-04-06
Updating python to version 3.8 in Dockerfile

alangsto:
## [MST-720](https://openedx.atlassian.net/browse/MST-720)

The push dockerfile step is now erroring because it can't find the python3.5-dev package. I'm upgrading the python version specific packages to 3.8 (mirroring what is done here: https://github.com/edx/course-discovery/blob/master/Dockerfile).


schenedx (approved):
👍

PR #407, opened 2021-04-15
[fix]: session cookie security

matthugs:
JIRA:EDUCATOR-5711


alangsto (approved):

matthugs:
not moving forward deploying this b/c it should be a no-op after the remote-config changes

PR #409, opened 2021-04-26
Added pin to social-auth-core for upgrade bot

alangsto:
## Description

This job (https://build.testeng.edx.org/job/registrar-upgrade-python-requirements/133/) was failing due to incompatible PYJWT version specified by different packages. I've added a pin on the social-auth-core version so that the job can run successfully.


requirements/constraints.txt
```diff
@@ -30,6 +30,8 @@ pylint==2.4.4
 analytics-python<2.0         # https://github.com/edx/registrar/pull/39
 celery<5.0                   # version 5.0 drops support for python 3.5
 edx-rest-api-client==1.9.2   # https://github.com/edx/registrar/pull/1
+social-auth-core<4.0.3       # higher versions require PYJWT>=2.0.0 which is
+                             # incompatible with the pinned edx-rest-api-client
```
matthugs, line 34: Why is edx-rest-api-client pinned, though?

Looks like it always has been for the whole existence of registrar, which is mysterious. Could we write a ticket in the MST backlog to upgrade that library?

matthugs, line 34: Why is edx-rest-api-client pinned, though?

Looks like it always has been for the whole existence of registrar, which is mysterious. Could we write a ticket in the MST backlog to upgrade that library?

matthugs (approved):

PR #415, opened 2021-06-02
Cleaned up constraints file and removed pin from edx-rest-api-client

alangsto:
## [MST-779](https://openedx.atlassian.net/browse/MST-779)

Removed pin for edx_rest_api_client and modified tests that failed as a result from the upgrade. 

I also went ahead and removed other constraints that I thought were unnecessary (due to python versions).


requirements/constraints.txt
```diff
@@ -17,23 +17,14 @@ djangorestframework==3.11.1
 # django-storages version 1.9 drops support for boto storage backend.
 django-storages<1.9
 
+# Pin to the latest release
+celery==5.0.4
 
-# Requires: Python >=3.6
```
alangsto, line 21: removed constraints that were in place due to python versions, as we now are using Python 3.8

requirements/constraints.txt
```diff
@@ -17,23 +17,14 @@ djangorestframework==3.11.1
 # django-storages version 1.9 drops support for boto storage backend.
 django-storages<1.9
 
+# Pin to the latest release
+celery==5.1.0
```
alangsto, line 21: pinned celery to a higher version because we are now using python 3.8

requirements/constraints.txt
```diff
@@ -17,23 +17,14 @@ djangorestframework==3.11.1
...
 
-# Requires: Python >=3.6
-mock<4.0.0
-zipp<2.0.0
+# higher versions require PYJWT>=2.0.0 which is incompatible with the common constraints file
+social-auth-core<4.0.3
```
alangsto, line 24: `social-auth-core` still has to be pinned because of a recent constraint that was added here: https://github.com/edx/edx-lint/blob/f45ce7e1a54a89f4e92a778d8411350e04b1ea71/edx_lint/files/common_constraints.txt#L25-L26

schenedx (approved):
👍

PR #417, opened 2021-06-03
Clean up constraints file and remove pin for edx_rest_api_client

alangsto:
## Description

The previous PR https://github.com/edx/registrar/pull/415 was reverted because the upgrade to celery was causing the registrar workers to go offline. 

I've left the pin on registrar but cleaned up a few other pins that are no longer needed.


schenedx (approved):
👍

PR #419, opened 2021-06-14
fix:add default transport so celery will run tasks eagerly

matthugs:
This makes sure we set a default transport, which is helpful when
we're running with CELERY_TASK_ALWAYS_EAGER. That setting enables us
to move forward without using the URL configured for the transport, so
this default doesn't matter when it's set. In other cases, this
will've needed to've been set explicitly before, so changing the
default should be quite benign.

Also I added a comment earlier to a Dockerfile but didn't get it into
somewhere, so tossing that in this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,7 @@ RUN apt-get install -qy \
 	libmysqlclient-dev \
 	libssl-dev
 RUN pip3 install --upgrade pip setuptools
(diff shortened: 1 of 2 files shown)


alangsto (approved):
💯

matthugs:
I should also be clear that this feels like a workaround for a celery bug, but I'm rollin' with it

PR #426, opened 2021-07-15
Adding django3.1 and 3.2 in tox for testing.

awais786:
Adding django3.1 and 3.2 in tox for testing. 
Run make upgrade to fetch latest packages.

Added allow failure for django3.2

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ on:
 jobs:
   run_tests:
     name: Tests
+    continue-on-error: true
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,7 @@ jobs:
     name: Tests
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os:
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -13,13 +13,17 @@ jobs:
     name: Tests
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
awais786, line 16: Run all tox-env to see all failure combinations failures.

registrar/apps/enrollments/apps.py
```diff
@@ -6,3 +6,4 @@
 
 class EnrollmentsConfig(AppConfig):  # pragma: no cover
     name = 'enrollments'
+    default = False
```
awais786, line 9: https://docs.djangoproject.com/en/3.2/ref/applications/#django.apps.AppConfig.default
```
Set this attribute to False to prevent Django from selecting a configuration class automatically. This is useful when apps.py defines only one AppConfig subclass but you don’t want Django to use it by default.
```

.github/workflows/ci.yml
mraarif: we can remove this since all the environments are passing now

mraarif (approved):

PR #427, opened 2021-07-16
feat: add group users to program and group model admin

alangsto:
## [MST-648](https://openedx.atlassian.net/browse/MST-648)

The list of users within a Program or Organization group is now easily visible within django admin.

Screen Shot 2021-07-16 at 9 14 29 AM

Screen Shot 2021-07-16 at 9 14 36 AM

View will automatically resize if there are many or longer usernames:

Screen Shot 2021-07-16 at 9 36 56 AM

Screen Shot 2021-07-16 at 9 37 12 AM


schenedx (approved):
👍

PR #444, opened 2021-10-14
style: fix string formatting due to pylint errors

bseverino:
## Description

Converted all `.format()` strings, and upgraded pylint + related packages. This should resolve the pylint errors found in https://github.com/edx/registrar/pull/443

### DIFF
--- a/registrar/apps/core/tests/factories.py
+++ b/registrar/apps/core/tests/factories.py
@@ -34,7 +34,7 @@ class GroupFactory(factory.django.DjangoModelFactory):
         model = Group
         django_get_or_create = ('name',)
 
-    name = factory.Sequence('group{}'.format)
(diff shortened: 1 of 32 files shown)


registrar/apps/core/tests/factories.py
```diff
@@ -34,7 +34,7 @@ class Meta:
         model = Group
         django_get_or_create = ('name',)
 
-    name = factory.Sequence('group{}'.format)
+    name = factory.Sequence('group{}'.format)  # pylint: disable=consider-using-f-string
```
matthugs: might be nice to do this one with a lambda as well, just for consistency? (I.e. similar to how the other sequence we needed to change in this file is instantiated)

matthugs (approved):
TIL about assertLogs, nice

PR #446, opened 2021-10-19
draft: testing github action outputs with generated PRs

matthugs:
## Description


matthugs:
alas, this doesn't work without output being on its own line https://github.com/edx/edx-proctoring/runs/3940725527?check_suite_focus=true

PR #449, opened 2021-11-22
feat: manage_programs now creates organizations if need be

matthugs:
This command is part of the build process for Master's Integration
Environments, the process of which is described here:
https://openedx.atlassian.net/wiki/spaces/PT/pages/1028489324/Master+s+Integration+Environment
The command did not previously create organizations but instead failed
when no authoring org.s were previously put in place by another
command. This is because Integration Environments writ large only
support a single authoring organization; we only authorize a user to
use one at a time. However, to have better advance visibility when we
wouldn't be able to create an Integration Environment in response to a
partner request, we'd like to add a wider variety of programs to our
auto-built canary environment.

## Description

### DIFF
--- a/registrar/apps/core/management/commands/manage_programs.py
+++ b/registrar/apps/core/management/commands/manage_programs.py
@@ -30,8 +30,7 @@ class Command(BaseCommand):
             discovery_details = ProgramDetails(uuidkey[0]).raw_data
             if not discovery_details:
                 raise CommandError('Could not read program from course-discovery; aborting')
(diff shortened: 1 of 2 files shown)


zacharis278 (approved):
so the goal of this is to create a bunch of orgs in the masters weekly box?

matthugs:
@Zacharis278 yes, the point is to avoid failing the build when we try to pull in programs from several org.s. We want to do that so we can catch a wider range of reasons why programs would fail to be pulled onto a sandbox. This is a change mostly motivated by ease of maintenance, and less so by narrowly tailoring this automation to a use case.

PR #458, opened 2022-02-16
chore: remove newrelic from registrar Dockerfile

arbabkhalil:
## Description

Moving newrelaic installation part and CMD for it to internal Dockerfile.


adzuci (approved):
:shipit:

PR #460, opened 2022-02-23
fix: Update the link of the confluence document detailing enrollment integration architecture

schenedx:
## Description

The link associated with the architecture decision of program enrollment storage to the confluence page should be updated to a confluence space which is visible to the public.


kdmccormick (approved):
Thanks again!

PR #462, opened 2022-03-10
fix: manually upgrade pip-tools

jansenk:
## Description

- upgrade pip-tools from 6.4.0 to 6.5.1
- add wrapt < 1.14.0 and responses < 0.17.0 to constraint.txt

If not, the upgrade does not pass testing


alangsto (approved):
👍

PR #464, opened 2022-03-15
chore: remove constraints related to db integrity errors

schenedx:
## Description
Remove the constraints on `moto` and `idna` because these can be address by unit test code changes.

### DIFF
--- a/requirements/common_constraints.txt
+++ b/requirements/common_constraints.txt
@@ -3,16 +3,6 @@
 # See BOM-2721 for more details.
 # Below is the copied and edited version of common_constraints
 
-# This is a temporary solution to override the real common_constraints.txt
-# In edx-lint, until the pyjwt constraint in edx-lint has been removed.
-# See BOM-2721 for more details.
-# Below is the copied and edited version of common_constraints
-
-# This is a temporary solution to override the real common_constraints.txt
-# In edx-lint, until the pyjwt constraint in edx-lint has been removed.
-# See BOM-2721 for more details.
-# Below is the copied and edited version of common_constraints
-
 # A central location for most common version constraints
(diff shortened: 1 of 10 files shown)


requirements/common_constraints.txt
```diff
@@ -13,6 +13,16 @@
...
+# This is a temporary solution to override the real common_constraints.txt
+# In edx-lint, until the pyjwt constraint in edx-lint has been removed.
+# See BOM-2721 for more details.
+# Below is the copied and edited version of common_constraints
+
+# This is a temporary solution to override the real common_constraints.txt
+# In edx-lint, until the pyjwt constraint in edx-lint has been removed.
+# See BOM-2721 for more details.
+# Below is the copied and edited version of common_constraints
+
```
matthugs: it looks like this comment is getting duplicatively added every time we run, e.g. `make upgrade`? We should maybe report that to BOM so they can help us avoid needing to think about it, and remove this from our diffs.

matthugs (changes requested):
Do we need `moto` outside of the test requirements? I wonder if we can just eliminate it from a number of requirements files (except for test) while we're looking at this?

schenedx:
@matthugs the `moto` library is only specified in the `test.in` file. However, `local.in`, `devstack.in` and `monitoring/requirements.in` all import `test.in`. That's why moto is installed on those. I don't want to expand the scope of this change to tease out requirements list dependencies.

requirements/common_constraints.txt
schenedx: Removed the duplicates from the PR. I'll investigate why `make upgrade` dupes this. That should be outside the scope of this PR tho.

alangsto (approved):
👍

PR #466, opened 2022-04-04
fix: remove newrelic from make file

ohnickmoy:
## Description
I noticed that this [github action failed](https://github.com/openedx/registrar/runs/5819036781?check_suite_focus=true) and I presume that because the new relic portion was removed in https://github.com/openedx/registrar/pull/458, then the references to new relic have to be removed in the make file as well. 


adzuci (approved):
LGTM

PR #468, opened 2022-04-06
fix: update dockerfile to use virtualenv and python 3.8

ohnickmoy:
## Description

Part of [PSRE-1460](https://openedx.atlassian.net/browse/PSRE-1460) and the work to containerize registrar. Adds lines to run setup within a virtual environment 

Updates the registrar docker file to use a virtualenv and python 3.8 instead of 3.5


Makefile
```diff
@@ -85,6 +85,7 @@ upgrade: piptools  $(COMMON_CONSTRAINTS_TXT)  ## re-compile requirements .txt fi
 	mv requirements/test.tmp requirements/test.txt
 
 piptools:
+	pip install -r requirements/pip.txt
```
adzuci, line 88: @ohnickmoy Can you describe why this was needed and/or how pip and setuptools was installed before this addition in places like devstack or production?

adzuci (approved):
I compared this Dockerfile to the one in https://github.com/openedx/edx-notes-api/blob/master/Dockerfile and think all these changes make sense.  It could be worth getting someone from #masters-cosmonauts to also review it before merging though, just in case they see something I miss.

P.S. please update the PR description.

Makefile
ohnickmoy: my rationale is based on what was described here: https://github.com/openedx/edx-notes-api/pull/206/files#r499761008 and here: https://github.com/edx/demographics/pull/69/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R36

I believe it's to pin pip and setup tools to a particular version

Dockerfile
```diff
@@ -39,8 +41,8 @@ WORKDIR /edx/app/registrar
 
 # Copy just Python requirements & install them.
 COPY requirements/ /edx/app/registrar/requirements/
-COPY Makefile /edx/app/registrar/
-RUN make production-requirements
+RUN pip install -r requirements/pip.txt
+RUN pip install -r requirements/production.txt
```
matthugs, line 45: Could you revert these lines? It looks like they do (very very nearly) the same thing, from reading the Makefile. (Also helps if the stuff called "production" stays the thing we do in production, and reverting these lines is the easiest way I can see of making sure that's still true in the future.)

matthugs (approved):

kdmccormick:

PR #498, opened 2022-11-17
APPEND_SLASH enabled and added logic to append slash

zubairshakoorarbisoft:
## Description

I've enabled APPEND_SLASH flag and added logic to append to forward slash in urls end where its absent.
Issue: https://github.com/edx/edx-arch-experiments/issues/79


kdmccormick:

PR #541, opened 2023-07-05
fix: Install pkg-config in image to satisfy mysqlclient==2.2.0

timmc-edx:
This requirement was recently upgraded, and was causing devstack provisioning tests to fail.


michaelroytman (approved):
Thank you!

PR #549, opened 2023-08-25
Fix django 4 deprecation warnings

UsamaSadiq:
## Description
Fixed `Django 4.0` and `Django 4.1` deprecation warnings. 
Created the PR under issue https://github.com/edx/edx-arch-experiments/issues/414.

### DIFF
--- a/registrar/apps/api/internal/urls.py
+++ b/registrar/apps/api/internal/urls.py
@@ -1,6 +1,6 @@
 """ API internal URLs. """
 
-from django.conf.urls import url
+from django.urls import re_path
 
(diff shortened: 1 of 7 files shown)


registrar/settings/base.py
```diff
@@ -179,8 +179,6 @@
 
 USE_I18N = True
 
-USE_L10N = True
```
awais786: Revert this it might break django32

awais786 (approved):

irtazaakram (approved):

registrar/settings/base.py
UsamaSadiq: reverted the change.

PR #550, opened 2023-08-28
Add Django 4.2 support in CI in Registrar

### LINKED ISSUE #208: [Registrar]: Django 4.2 Upgrade
## Description
Under the effort of [Django 4.2 Upgrade](https://github.com/openedx/public-engineering/issues/199), complete all of the following steps to complete the upgrade. 

- [x] Update tox & Github action workflow using modernisers to add support for `Django 4.2`
- [x] Remove any versions of Python earlier than 3.8 from `tox.ini`, and `GitHub Actions` workflows.
- [x] Update the pinned `Django` version in the requirements to `Django==4.2`
- [x] Run `make upgrade` to update all dependencies for `Django 4.2`.
- [x] Run and verify all tests are passing in the CI for both `Django 4.2` and `Django 3.2` to contain backward compatibility.
- [x] Run available code-mods to fix the failing tests. 
- [ ] Add conditional checks wherever needed to still have support for `Django 3.2`. 
- [ ] Update the repo support field in the [IDA Upgrade Sheet](https://openedx.atlassian.net/wiki/spaces/AC/pages/3815637011/Django+4.2+Upgrade+List+of+IDAs).

irtazaakram:
## Description

- Resolves https://github.com/openedx/public-engineering/issues/208


registrar/apps/api/v3/views.py
```diff
@@ -20,24 +19,7 @@
 
 logger = logging.getLogger(__name__)
 
-SCHEMA_COMMON_RESPONSES = {
```
UsamaSadiq: Please mention why this change is needed?

irtazaakram: https://github.com/openedx/registrar/pull/552 
This PR has explanation for these changes. It should b merged before this PR.

tox.ini
```diff
@@ -1,13 +1,16 @@
...
     django32: -r requirements/django.txt
+    django42: Django>=4.2,<5.0
     -r{toxinidir}/requirements/test.txt
 whitelist_externals =
     i18n_tool
     /bin/bash
-commands = {posargs:pytest}
\ No newline at end of file
+commands = {posargs:pytest}
+setenv =
```
awais786, line 15: Tox gives error locally due to boto issue.

registrar/apps/core/tests/test_discovery_cache.py
```diff
@@ -122,14 +122,23 @@ def test_discovery_program_get(self, disco_status, disco_json, expected_raw_data
         assert isinstance(loaded_program, ProgramDetails)
         assert loaded_program.uuid == self.program_uuid
         assert loaded_program.raw_data == expected_raw_data
-        self.assertEqual(len(responses.calls), 1)
+
+        # due to cache issue, it fails when run it as single test.
+
+        self.assertIn(
```
zacharis278: this doesn't seem to be testing the same thing anymore? It looks to me this test was previously ensuring the cache works instead of making multiple requests to discovery.

awais786: @Zacharis278 there is problem in this test file. It does not fulfill the cache itself it is depending on some other test file. 

This `pytest registrar/apps/core/tests/test_discovery_cache.py` fails.
but it is working 
```
pytest  registrar/apps/core/tests/test_api_client.py registrar/apps/core/tests/test_discovery_cache.py. 
```
test_api_client fulfills the cache with login attempt.

Idealy `test_discovery_cache.py` can add new step or test which can warm the cache itself.

awais786: https://docs.djangoproject.com/en/4.1/topics/testing/tools/#testcase it might be related. thats why it works on django32 ci but not on django42.

awais786:
@Zacharis278 you can review this again.

registrar/apps/core/tests/test_discovery_cache.py
zacharis278: ah so the request for login is what's messing up the assertion here?

Would this fix the test:

- keep the assert you have on 128
- reset the responses calls or keep track of the number. We don't really care how many requests happen here I think so the original assert was a bit brittle to how login works.
- on line 138 make sure there are no new responses calls since we shouldn't be be making a network request for the second ProgramDetails object.

awais786: Import is enough here. Since that test file is doing all magic. Owning team can fix it in future in a proper may.
Above mentioned steps does not fix the test, it that need more time.

awais786: If this is not an ideal solution i can give a try to your suggestion as well.

UsamaSadiq (approved):

registrar/apps/core/tests/test_discovery_cache.py
```diff
@@ -14,6 +14,8 @@
 
 from ..api_client import DISCOVERY_API_TPL, DiscoveryServiceClient
 from ..discovery_cache import ProgramDetails
+from ..tests.test_api_client import \
```
UsamaSadiq, line 17: Why is this change needed?

PR #551, opened 2023-08-29
feat!: upgrading django-storages

irtazaakram:
This Package needs to upgrade to get django42 support.

## Description

More details on this can be found here https://github.com/openedx/edx-platform/pull/32571


awais786:
https://github.com/edx/edx-internal/blob/master/argocd/applications/registrar/prod-config.yaml#L95
as per my understanding objects are saving as private in bucket. because null overrides the default `acl` set by package.
So this package upgrade will make no effect on acl behaviour.

awais786 (approved):

zacharis278 (approved):

PR #563, opened 2023-10-03
fix: add conditional trusted origins update

UsamaSadiq:
## Description
- PR created under the issue https://github.com/edx/edx-arch-experiments/issues/459
- Adding conditional approach to use CSRF_TRUSTED_ORIGINS_WITH_SCHEME for Django>4


iamsobanjaved (approved):

awais786:
https://github.com/edx/edx-internal/pull/9329/files#diff-312f6369d6d9b20e80b359157dc6f65c9cdab0e50aaf2bf7d2c5b17847cac701R56

Settings already merged in internal PR.

zacharis278 (approved):
huh looks like codecov is no longer ignoring the registrar/settings folder which is causing the check to fail on this.

Might be an older issue, just wanted to point it out.

PR #565, opened 2023-10-05
chore: `test_discovery_program_get` this test fails locally

awais786:
**using django32**:
When run test_discovery_program_get locally it fails due to cache issue.
```
pytest registrar/apps/core/tests/test_discovery_cache.py::ProgramDetailsTestCase::test_discovery_program_get_1
AssertionError: 2 != 1
```

Need to fix since its failing on `django42` ci also.
https://github.com/openedx/registrar/pull/550 its green after cherrypick.


zacharis278:
left a comment in https://github.com/openedx/registrar/pull/550

PR #584, opened 2023-12-05
feat: add success log to disco sync

zacharis278:
Log a message when this job succeeds. We will use this to keep track of job successes for the purposes of alerting.

Right now we log when each component of this succeeds but not the job as a whole. This is an single point of truth to report on

2U internal alert policy: https://onenr.io/0bRmZ3gJWjy


ilee2u (approved):

schenedx (approved):
:+1:

PR #596, opened 2024-02-14
feat: handle nonslug program keys from discovery

zacharis278:
## Description

This [Discovery ADR](https://github.com/openedx/course-discovery/blob/ef5f9dcb3713d99d8ebf081346c29d1b72bb80e7/docs/decisions/0023-url-restructuring-for-seo-improvements.rst) changed the way Program marketing slugs are generated such that they are no longer guaranteed to be a valid SlugField which makes them unsafe to use as url tokens. Because Registrar supports these values deviating from the source in discovery we're able to translate any non-slug values to something that will work with Registrar and any UI's that rely on it (Program Console + Learner Portal).

This change will only impact new programs moving forward since we do not modify a programs key after a record with that UUID has been synced. This is done to support custom external keys for partners that need a specific value and works to our benefit here.

### DIFF
--- a/registrar/apps/core/management/commands/sync_with_discovery.py
+++ b/registrar/apps/core/management/commands/sync_with_discovery.py
@@ -3,6 +3,7 @@ import logging
 
 from django.core.management.base import BaseCommand
 from django.db import transaction
+from slugify import slugify
 
 from registrar.apps.core.api_client import DiscoveryServiceClient
 from registrar.apps.core.models import (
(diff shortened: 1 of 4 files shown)


registrar/apps/core/management/commands/sync_with_discovery.py
```diff
@@ -127,14 +128,16 @@ def sync_programs(self, program_types):
                 if discovery_authoring_org:
                     auth_org = existing_org_dictionary.get(discovery_authoring_org.get('uuid'))
                 if auth_org:
+                    # key fromd disco is not guarenteed to be a valid slugfield
```
michaelroytman, line 131:
```suggestion
                    # key from disco is not guaranteed to be a valid slugfield
```

michaelroytman (approved):

schenedx (approved):
👍

PR #600, opened 2024-02-27
chore: add python slugify to prod

varshamenon4:
Add python-slugify to production so that sync_to_discovery management command works.

### DIFF
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -19,6 +19,7 @@ edx-auth-backends
 edx-django-release-util
 edx-drf-extensions
 edx-rest-api-client
+python-slugify
 pytz
 redis
 social-auth-app-django
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -161,6 +161,10 @@ pynacl==1.5.0
     # via edx-django-utils
 python-dateutil==2.8.2
     # via analytics-python
+python-slugify==4.0.1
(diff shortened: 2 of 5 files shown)


requirements/common_constraints.txt
```diff
@@ -3,6 +3,11 @@
 # See BOM-2721 for more details.
 # Below is the copied and edited version of common_constraints
 
+# This is a temporary solution to override the real common_constraints.txt
```
michaelroytman: It seems this paragraph was duplicated. Did you use `make upgrade`? I think we could probably remove this paragraph, right?

alangsto (approved):

rijuma (approved):

michaelroytman (commented):
There are a lot of other dependency upgrades in here. I feel like that can get a little cluttered to review outside of a Python requirements upgrade pull request. Are you able to just include the installation of `python-slugify`? I usually manually modify the `upgrade` Make target to remove the `--upgrade` flag to avoid upgrading all the dependencies. But let me know if you think the pain isn't worth it.

requirements/common_constraints.txt
varshamenon4: I like your suggestion above to remove all extraneous upgrades, hopefully that will remove this as well! I'll do that!

PR #622, opened 2024-06-13
fix: constrain requests package version

ilee2u:
A quick PR to constrain the `requests` package to >=2.32.3 so that only one version of the package is installed, as not constraining this caused an error where versions 2.31.0 and 2.32.3 would also be installed at the same time, causing a version conflict.

See the changes made to `constraints.txt` for more details.


varshamenon4 (approved):
Nice find!

PR #624, opened 2024-06-14
fix: import base to docs reqs

ilee2u:
An improvement on this PR: https://github.com/openedx/registrar/pull/622

Previously, a version conflict between v2.31.0 and v2.32.3 of the `requests` package occurred when attempting to do python requirement upgrades. The above PR implemented a hotfix by pinning the requests version >=2.32.3, though we decided that this isn't sustainable.

The real problem seemed to be that `docs.in` never inherited requirements from `base.in`, causing the docs requirements to not account for what version of requests was being downloaded by other requirement files. This PR fixes that.


varshamenon4 (approved):
lgtm!

PR #633, opened 2024-07-23
docs: add 2u-cosmonauts as owner

zacharis278:
https://open-edx-proposals.readthedocs.io/en/latest/processes/oep-0055/decisions/0001-use-backstage-to-support-maintainers.html

If 2U is interested in owning program-console we should also own registrar. To confirm before merging.


rijuma (approved):

schenedx (approved):
👍

PR #643, opened 2024-09-04
perf: clean docker files

huniafatima-99:
## Description

This PR is the part of [devstack DEPR ticket](https://github.com/edx/edx-arch-experiments/issues/783). In this PR we are removing Dockerfile from registrar IDA


kdmccormick:

